PR #1, opened 2020-05-12
Removing static files

TysonRV:
- Removing a good bunch of clearly unused static files.


farridav:
Looks good, this will reduce the size of the package significantly :)

Did you have loo at all the core views to ensure everything still works as expected ? No 404s in the network panel ?

farridav:
LGTM :+1:

TysonRV:
@farridav Yeah, all good.

PR #9, opened 2020-05-12
Upgrade font awesome to 5

TysonRV:
- Upgrade FontAwesome from version 4 to 5 using CDN.

References for the icons:
https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#name-changes


farridav:
CDN is good, we have other stuff going via CDN too, it will keep the size of the package down ... why are there a shit load of files ? i though it was just 1 css and 1 svg?

TysonRV:
@farridav Yeah, didn't need some of the stuff in there... Anyway, I swapped it already as it was just too many files

farridav:
needs a rebase (my fault) then looks good

PR #17, opened 2020-05-16
List view

farridav:


jazzmin/templates/admin/change_form.html
```diff
@@ -6,139 +6,77 @@
...
-    </form>
-</div>
-
-
-{% endblock %}
-
-{% block extrajs %}
-<script src="{% static "adminlte/plugins/datatables/jquery.dataTables.min.js" %}"></script>
-<script src="{% static "adminlte/plugins/datatables/dataTables.bootstrap.min.js" %}"></script>
+    </form></div>
```
TysonRV, line 81: If this is just formatting (and the BC), we could use my version instead.

TysonRV: Formatting -> getting the django admin site 😄

PR #18, opened 2020-05-16
Add & change views

TysonRV:
- Improvements made to add & change views, avoiding touching CSS yet.
- I've changed some buttons as well, for `outline` ones. It looks better IMO, but open to suggestions!


farridav:
Nice, lets get this in, and follow up with some more improvements individually !

PR #19, opened 2020-05-17
Adapting password form template

JuanVT:
- Adding classes to breadcrumb.


farridav:
This is fine to go in as is, no need for WIP for 2 days

PR #20, opened 2020-05-17
List view improvements

farridav:
- Get list view filters to work properly with select2
- Improve appearance of date heirarchy
- Show when list tables are sorted (were not using datatables, as there is no need)
- Fix paginator


jazzmin/templatetags/jazzmin.py
```diff
@@ -192,5 +192,4 @@ def debug(value):
     """
     Add in a breakpoint here and use filter in templates for debugging ;)
     """
-    import ipdb; ipdb.set_trace()
```
TysonRV, line 195: I think this is exactly what this function is for.. if you remove the BP, won't do what is for? Anyway, I added a #NOQA tag in my branch to this, to prevent flakes.

farridav: lol it is what its for, but im removing it as leaving it in would mean we need to add ipython and ipdb with the package

TysonRV: oh gotcha! 👍 I guess we can get rid of the function then

PR #21, opened 2020-05-18
Bits and bobs

farridav:
- Makes the history component on the dashboard extra jazzy
- default to adminlte logo not that bird thing
- use sensible avatar default
- delete unused images
- link user email to profile
- Fix dashboard link
- Mark active page with JS
- Make side menu smaller and more compact


TysonRV:
This all looks good to me :shipit:

PR #34, opened 2020-05-22
bits and bobs

farridav:


jazzmin/static/jazzmin/img/calendar-icons.svg
```diff
@@ -0,0 +1,14 @@
...
+    </g>
+    <g id="next">
+      <path d="M845 1395l454-454q19-19 19-45t-19-45l-454-454q-19-19-45-19t-45 19l-102 102q-19 19-19 45t19 45l307 307-307 307q-19 19-19 45t19 45l102 102q19 19 45 19t45-19zm819-499q0 209-103 385.5t-279.5 279.5-385.5 103-385.5-103-279.5-279.5-103-385.5 103-385.5 279.5-279.5 385.5-103 385.5 103 279.5 279.5 103 385.5z"/>
+    </g>
+  </defs>
+  <use xlink:href="#previous" x="0" y="0" fill="#333333" />
+  <use xlink:href="#previous" x="0" y="1792" fill="#000000" />
+  <use xlink:href="#next" x="0" y="3584" fill="#333333" />
+  <use xlink:href="#next" x="0" y="5376" fill="#000000" />
+</svg>
```
TysonRV, line 14: Is this exactly the same file coming from Django? I've learnt that even if they aren't there in the project, they will be as long as the project has Django installed on it, so it is kind of redundant as jazzmin will live always in a Django project (obviously, in the case they are exactly the same).

jazzmin/static/jazzmin/js/main.js
```diff
@@ -1,26 +1,55 @@
...
 }
 
-$(document).ready(function() {
-    $('a[href="' + window.location.pathname + '"]').addClass('active');
-
-    $('.actions select').addClass('form-control');
+function eraseCookie(key) {
+    var keyValue = getCookie(key);
+    setCookie(key, keyValue, '-1');
+}
```
TysonRV, line 15: Where is this being used?

jazzmin/static/jazzmin/img/selector-icons.svg
```diff
@@ -0,0 +1,34 @@
...
+  <use xlink:href="#down" x="0" y="5376" fill="#447e9b" />
+  <use xlink:href="#left" x="0" y="7168" fill="#666666" />
+  <use xlink:href="#left" x="0" y="8960" fill="#447e9b" />
+  <use xlink:href="#right" x="0" y="10752" fill="#666666" />
+  <use xlink:href="#right" x="0" y="12544" fill="#447e9b" />
+  <use xlink:href="#clearall" x="0" y="14336" fill="#666666" />
+  <use xlink:href="#clearall" x="0" y="16128" fill="#447e9b" />
+  <use xlink:href="#chooseall" x="0" y="17920" fill="#666666" />
+  <use xlink:href="#chooseall" x="0" y="19712" fill="#447e9b" />
+</svg>
```
TysonRV, line 34: This file I added and removed after realising it was also implicit in Django (as I was basically copying it over), thinking on keeping the framework as lightweight as possible

farridav: The icons were not rendering out, the SVG was referenced from our app.css, with a relative path, so it wasn't finding the admin one, as that lives under admin/img/ and our app.css will be looking in jazzmin/img/ 

If you can find a way to get the multi select in the user detail view to show the arrows without this asset, we can remove it :)

TysonRV: Mm... interesting.. I was pretty sure it was doing it for me. Maybe we are talking about different ones, but I will investigate about it.

PR #16, opened 2020-05-14
Upgrade to adminlte3

farridav:
Upgrade to AdminLTE3, and upgrade all of its dependencies, basically re-write all templates, and run through entire admin

## TODO
- [x] Run test app in supported Django versions, provide compatibility fixes
- [x] Add supported python and Django versions to README [example [here](https://shields.io/)]
- [x] Cleanup unused static files
- [x] run through main breakpoints (Small, Medium & large screen sizes) and fix where necessary
- [x] Make new screenshots

PR #36, opened 2020-05-24
Datetime fields in changeform

TysonRV:
- Giving some color to datetime fields in change_form

Far from perfect... but at least as a first version of it. 

Screenshot 2020-05-28 at 21 23 47

Screenshot 2020-05-28 at 21 30 03

### DIFF
--- a/jazzmin/static/jazzmin/js/change_form.js
+++ b/jazzmin/static/jazzmin/js/change_form.js
@@ -1,0 +1,9 @@
+$(document).ready(function() {
+    $('.vDateField').addClass('form-control');
+
+    <div class="input-group-prepend">
+                      <span class="input-group-text">
+                        <i class="far fa-calendar-alt"></i>
+                      </span>
+                    </div>
+});
--- a/jazzmin/static/jazzmin/js/change_form.js
+++ b/jazzmin/static/jazzmin/js/change_form.js
@@ -1,4 +1,4 @@
-function prependCalendarIcon() {
+function jazzDatetimeFields() {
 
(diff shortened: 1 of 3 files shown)


farridav:
Needs a rebase ;) make sure your not doing things that are already done , see https://github.com/farridav/django-jazzmin/pull/16 for up to date list of things that need sorting

jazzmin/static/jazzmin/js/change_form.js
```diff
@@ -0,0 +1,34 @@
...
+            '</div></p>'
+        )
+    });
+}
+
+function jazzIcons() {
+    setTimeout(function(){
+        $(".date-icon").attr("class", "fa fa-calendar");
+        $(".clock-icon").attr("class", "fa fa-clock");
+    }, 500);
```
TysonRV, line 28: This is mostly me testing the icons, but that's doable with CSS

farridav: SetTimeout .. that's nasty

TysonRV: Ha! yeah, we don't want that...

farridav:
Are you happy with the final implementation ?

TysonRV:
@farridav only for now. I would rather move on to something else and get this improved as a further step or second version. We still have to deal with making the whole thing responsive, so I'm sure more ideas will flow in :) 

Also, the most critical point are those timeouts.. but I haven't found out how to get rid of them and get the implementation I wanted without getting all full js-mode. Open to suggestions!

PR #44, opened 2020-06-06
Styling docs pages

TysonRV:
Jazzin'up pages on `/admin/doc/` 

- [x] Index page
- [x] Tags
- [x] Filters
- [x] Models
- [x] Model pages
- [x] Views
- [x] Bookmarklets


farridav:
Looking good so far ;) I wonder whether you could use vertical tabs for any of these pages, as in https://adminlte.io/themes/v3/pages/UI/navbar.html

TysonRV:
@farridav This is pretty much done for now. I ended up using `accordeons` for most of the stuff, but made it easy to swap in case it is just too much blue. Also, models has a different layout, I might change it to more `accordeons` to keep it consistent. 

Anyway, all good to go for now, with room for improvement 🧀

farridav:
This looks jazzy as F, im merging it in, we will follow up with screenshots

PR #46, opened 2020-06-09
Bits & Bobs

farridav:
- Ensure we dont allow collapse class in fieldsets (As django admin js stops it rendering out the content)
- Allow user to add in custom CSS or JS (using a relative path)

### DIFF
--- a/jazzmin/templates/admin/includes/fieldset.html
+++ b/jazzmin/templates/admin/includes/fieldset.html
@@ -1,5 +1,5 @@
 {% load jazzmin %}
-<div class="card {% safe_classes fieldset.classes %}">
+<div class="card {% strip_collapse_class fieldset.classes %}">
     <div class="p-5{% if fieldset.name %} card-body{% endif %}">
         {% for line in fieldset %}
         <div class="form-group{% if line.fields|length_is:'1' and line.errors %} errors{% endif %}{% if not line.has_visible_field %} hidden{% endif %}{% for field in line %}{% if field.field.name %} field-{{ field.field.name }}{% endif %}{% endfor %}">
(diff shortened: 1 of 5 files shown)


jazzmin/templates/admin/includes/fieldset.html
```diff
@@ -1,4 +1,5 @@
-<fieldset class="card {{ fieldset.classes }}">
+{% load jazzmin %}
+<div class="card {%  safe_classes fieldset.classes %}">
```
TysonRV, line 2:
```suggestion
<div class="card {% safe_classes fieldset.classes %}">
```

jazzmin/templatetags/jazzmin.py
```diff
@@ -265,3 +265,15 @@ def header_class(header, forloop):
...
+@register.simple_tag()
+def safe_classes(classes):
+    """
+    Remove reserved classes from CSS class list
+    """
+    if type(classes) == str:
+        classes = classes.split(' ')
+
+    excluded = ['collapse']
+    return ' '.join([x for x in classes if x not in excluded])
```
TysonRV, line 279: Maybe, if it is just a string, you could use `replace` instead for readability. It depends whether you want to add more stuff to `excluded` in the future or not.. TBH, I would call that tag just `strip_collapse_class`, as that's what it is really doing. 🙂 

```
@register.simple_tag()
def strip_collapse_class(classes):
    """
    Remove collapse class from CSS class list
    """
    if type(classes) == str:
        classes = classes.replace('collapse', '')

    return classes
```

It looks like it can only be a string, so that would suffice. Anyway, cheese for thought! 🧀

farridav: Good shout, you drove me to a much cleaner solution, https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#cut

TysonRV: That's just brilliant

Issue #49, opened 2020-06-11
Responsive UI


neowenshun:
Hello thanks so much for your hardwork , it has made my django admin look more modern and 'jazzy'. Apart from that , i have been fiddling around with it and realize that certain portions of the template is not responsive , for example , the breadcrumb will overlap with the 'dashboard' title when changed to mobile mode and in the 'ipad' mode the actions menu seems cluttered. Is there a way to override these?

farridav:
Yes you can override with custom CSS/JS (See https://django-jazzmin.readthedocs.io/configuration/#ui-tweaks)

Though if there is a change you want to put up a small PR for, we can get it in the project (See https://django-jazzmin.readthedocs.io/development/ for getting setup and contributing)

PR #52, opened 2020-06-12
Responsive tweaks

farridav:
Go through all breakpoints again, 

- Use `col-12` or `col-auto` as starting size
- Use `col-md-x` or `col-lg-x` as content scales up and looks decent
- Fix class names without qoutes
- Removes username from user menu, just uses icon now

Aims to fix https://github.com/farridav/django-jazzmin/issues/49


TysonRV:
Awesome, looks great now. Shipping!

PR #56, opened 2020-06-13
Making history page look a bit jazzier

TysonRV:
- Adding some color and icons to the history page to make it look less dull


farridav:
Nice work !!

PR #66, opened 2020-06-18
Fix multiple form fields in one line

RamezIssac:
Hello !

Thanks for the wonderful work

This PR should make multiple form fields in one form line display nicely. 
https://imgshare.io/images/2020/06/18/Selection_306.png

_ps: in line 14-18, that big if... I hope i didn't mess the `checkcard-row` case there :)_


farridav:
LGTM, the test are failing because it cant report coverage, ill resolve this seperately

PR #68, opened 2020-06-19
Enhancements

RamezIssac:
* Added a block surrounding the sidebar for easier customization 
* Moved code from base_site to base.html. Because base_site.html maybe overridden and without this change, the dev would need to copy the page_content block in _their_ base_site template.
* Added block.super in change_list extrajs


jazzmin/templates/admin/base_site.html
```diff
@@ -3,65 +3,16 @@
...
 {% endblock %}
 
 {% block footer %}
-    <div class="float-right d-none d-sm-inline">
-        <b>Jazzmin Version</b> {% get_jazzmin_version %}
-    </div>
-
-    <strong>Copyright &copy; {% now 'Y' %} {{ jazzmin_settings.copyright }}.</strong> All rights reserved.
+    {{ block.super }}
 {% endblock %}
```
TysonRV, line 16: If you are only going to leave `{{ block.super }}` in your `{% block %}`, you could just remove it all

TysonRV:
Thanks for this @RamezIssac 👍 
I think this kind of makes sense, as it will provide more flexibility to override the `base.html`. Basically, we keep `base_site.html` completely clean as an island for applications specific changes. This means: @RamezIssac you can get rid of everything and just leave:

```
{% extends 'admin/base.html' %}
```

Let's see if @farridav has something to add

jazzmin/templates/admin/base_site.html
RamezIssac: I totally agree.. left there for demonstrating purposes not more. It's cleaner to have just the extend line

TysonRV: Sounds good to me 🙂

farridav:
Seems sane to me, though I'd like to check it out locally and have a quick run through, I should get to that today hopefully.

farridav:
Merge when ready @TysonRV ill try to get the coveralls publishing to stop failing PRs

PR #69, opened 2020-06-22
pre-fill ui tweaks from user config

farridav:
- When a user uses `show_ui_builder: True` setup the builder with their current customisations so they can pick up where they left off..
- Fix duplicate named admin class
- put JS into functions

### DIFF
--- a/tests/test_app/polls/admin.py
+++ b/tests/test_app/polls/admin.py
@@ -79,13 +79,13 @@ class LogEntryAdmin(admin.ModelAdmin):
 
 
 @admin.register(Cheese)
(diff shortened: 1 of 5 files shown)


tests/test_app/polls/admin.py
```diff
@@ -79,13 +79,13 @@ def modified(self, obj):
 
 
 @admin.register(Cheese)
-class CampaignAdmin(admin.ModelAdmin):
+class CheeseAdmin(admin.ModelAdmin):
```
TysonRV, line 82: Oops... my fault

Issue #73, opened 2020-06-24
History page breaks when using a custom user model 


farridav:
This is because we have a url reversal to `auth_user` ..

PR #74, opened 2020-06-24
Dont reverse auth user model, as some users have custom user model

farridav:
Ensure we lookup urls based on the object, to cater for swapped out/custom  user models

Fixes https://github.com/farridav/django-jazzmin/issues/73

### DIFF
--- a/jazzmin/templates/admin/object_history.html
+++ b/jazzmin/templates/admin/object_history.html
@@ -39,7 +39,7 @@
                                     <i class="fas fa-{{ action_message|get_action_icon }} bg-{{ action_message|get_action_color }}"></i>
                                     <div class="timeline-item">
                                       <h3 class="timeline-header no-border">
-                                          <a href="{% url 'admin:auth_user_change' action.user.id %}" target="_blank">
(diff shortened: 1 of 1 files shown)


jazzmin/templates/admin/object_history.html
```diff
@@ -39,7 +39,7 @@ <h4 class="card-title">
                                     <i class="fas fa-{{ action_message|get_action_icon }} bg-{{ action_message|get_action_color }}"></i>
                                     <div class="timeline-item">
                                       <h3 class="timeline-header no-border">
-                                          <a href="{% url 'admin:auth_user_change' action.user.id %}" target="_blank">
+                                          <a href="{% jazzy_admin_url action.user %}" target="_blank">
```
TysonRV, line 42: Probably you noticed already, but it should be:

```
{{ action.user|jazzy_admin_url }}
```

farridav: dang its a filter

farridav: Loving our test coverage right now

PR #85, opened 2020-07-13
Minor tweak to top menu naming

mcclory:
The name of the top menu currently uses the AppConfig `name` property. This edit changes how that property is accessed so that it uses the `verbose_name` field if it's supplied and defaults to `name` as the default.


farridav:
Looks good, it's on me to fix the failing coveralls command

Issue #86, opened 2020-07-15
Add option to skip tabs for InlineModelAdmin


RamezIssac:
Hello !

I propose an enhancement that adds an option to jazzmin settings to skip creating the tabs for inline formsets.. and display them flat in the same page. An invoice is a typical use case for this.

Ideally this should be per each InlineModelAdmin , but as this can be tricky, let's start with a global On/Off option.

I'll submit a PR.

PR #87, opened 2020-07-15
Adds option `display_inlines_in_tabs`

RamezIssac:
Addresses the idea of  #86

### DIFF
--- a/jazzmin/settings.py
+++ b/jazzmin/settings.py
@@ -51,6 +51,7 @@ DEFAULT_SETTINGS = {
     # Icons that are used when one is not manually specified
     "default_icon_parents": "fa-chevron-circle-right",
     "default_icon_children": "fa-circle",
+    "display_inlines_in_tabs": True,
     #############
     # UI Tweaks #
     #############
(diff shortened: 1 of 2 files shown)


farridav:
Unfortunately this does not have the desired effect, using this option will stop tabs being used for fieldsets too, not just for the inlines.

If we want to have a setting which is `"changeview_format": "tabs|single"` then it would reflect this behaviour, though from looking at how `"single"` would render out in the demo app, we would need to add a little bit of jazz on the fieldsets within the form, as currently its 1 flat page, with no real distinction between the fieldsets, also, the setting would be global, so users would not have per model control over the rendering, for this, we could add in overrides with `"changeview_format_overrides": {"auth,user": "single", "auth.group": "tabs"}`.

Given this is scope creep on what you were trying to achieve, but does allow you to get there, whilst maintaining the flexibility we strive for, i can take a look at implementing this over the next couple of days, and you can let me know whether it gives you the desired outcome.

farridav:
Closing in favour of https://github.com/farridav/django-jazzmin/pull/88, please comment on that PR if it doesnt meet requirements

Issue #92, opened 2020-08-01
Carousel collapsible horizontal_tabs vertical_tabs does not work properly


fengyouchao:
`carousel` `collapsible` `horizontal_tabs` `vertical_tabs` does not work properly when `settings.LANGUAGE_CODE` is set to Chinese, Japanese and other non-Latin alphabet languages. `slugify` filter  does not support unicode.

PR #93, opened 2020-08-01
Use unicode_slugify instead of slugify

fengyouchao:
Fix the bug in #92

### DIFF
--- a/jazzmin/templatetags/jazzmin.py
+++ b/jazzmin/templatetags/jazzmin.py
@@ -2,6 +2,7 @@ import copy
 import itertools
 import json
 import logging
+import unidecode
 import urllib.parse
 
 from django.conf import settings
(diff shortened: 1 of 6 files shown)


fengyouchao:
The user menu is only displayed when the user is a super administrator. This is caused by a logic error in the `can_view_self` filter. I have fixed the bug in e351173

farridav:
@TysonRV can you take point on this ?

jazzmin/templatetags/jazzmin.py
```diff
@@ -417,3 +418,8 @@ def style_bold_first_word(message: str) -> str:
     message = " ".join([word for word in message_words])
 
     return format_html(message)
+
+
+@register.filter
+def unicode_slugify(message: str) -> str:
+    return defaultfilters.slugify(unidecode.unidecode(message))
```
TysonRV, line 425: hi @fengyouchao ! First of all, thanks for the hard work finding the issue and getting this to work! 🥳 


I think we could simplify that a bit and use: 

`slugify(message, allow_unicode=True)`

That way we wouldn't have to use unidecode I think. I'll perform the changes, but could you double check just in case I'm not reproducing the bug in the same way you are?

Thanks a lot for that!

TysonRV:
Closing this as it is on #95

PR #95, opened 2020-08-04
Use of unicode slugs & Fixes error in how to display Users application

### LINKED ISSUE #92: Carousel collapsible horizontal_tabs vertical_tabs does not work properly

TysonRV:
Resumes work on #93 

- Fixes #92
- Fixes error with User menu not showing unless logged user is superuser

Thanks again @fengyouchao


farridav:
Nice work

PR #98, opened 2020-08-05
Fix usermenu

fengyouchao:
Users without view_user permission cannot see the usermenu, which makes these users unable to log out or change their passwords. This bug first appeared in a0cca8b


farridav:
Looks good to me :) nice spot

Issue #99, opened 2020-08-05
apps not shown


navi380:
I have 5 apps installed in my project which are working fine for me in my Django admin. But i installed jazzmin admin everything is working fine for now but its shows me only 3 apps. further my other 2 apps name start with capital letter. i tried to rename them and check it again but it didn't resolve my issue.

farridav:
Can you post your jazzmin config here ? We will try to repro.

Screenshots can also be a great help if possible

navi380:
I was sure that may be this was name problem..  Cause only two apps have
name start with capital letters  like Order and Reviews..  And that are not
showing  in admin side..
So i chang the names by refactor but thats doesn't  work so  delete those
apps and again start new apps with small letters and did copy the code
there now its working  fine for me..  Its showing  me both of them as well

navi380:
`[](url)`

navi380:
'jazzmin',
 'django.contrib.admin',


JAZZMIN_SETTINGS = {
    # title of the window
    'site_title': 'shopna Admin',

    # Title on the brand, and the login screen (19 chars max)
    'site_header': 'Shopna',

    # square logo to use for your site, must be present in static files, used for favicon and brand on top left

    # Welcome text on the login screen
    'welcome_sign': 'Welcome to Shopna Admin',

    # Copyright on the footer
    'copyright': 'Acme Ltd',

    # The model admin to search from the search bar, search bar omitted if excluded
    'search_model': 'auth.User',

    # Field name on user model that contains avatar image
    'user_avatar': None,

    ############
    # Top Menu #
    ############

    # Links to put along the top menu
    'topmenu_links': [

        # Url that gets reversed (Permissions can be added)
        {'name': 'Home',  'url': 'admin:index', 'permissions': ['auth.view_user']},


        # model admin to link to (Permissions checked against model)
        {'model': 'auth.User'},


    ],

    #############
    # User Menu #
    #############



    #############
    # Side Menu #
    #############

    # Whether to display the side menu
    'show_sidebar': True,

    # Whether to aut expand the menu
    'navigation_expanded': True,

    # Hide these apps when generating side menu e.g (auth)
    'hide_apps': [],

    # Hide these models when generating side menu (e.g auth.user)
    'hide_models': [],

    # List of apps to base side menu ordering off of (does not need to contain all apps)
    'order_with_respect_to': ['accounts', 'polls',],

    # Custom links to append to app groups, keyed on app name
    'custom_links': {
[...]

PR #100, opened 2020-08-05
Allow for upper case app names

farridav:
Because we lower case app names during menu creation, but didnt lower case permission codenames, apps with upper case letters in their name would not be rendered out.

This is an oversight, and is fixed by this PR :)

Fixes https://github.com/farridav/django-jazzmin/issues/99 

Thanks for spotting this @navi380

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,4 +1,5 @@
 import pytest
+from django.db.models.functions import Upper
 from django.urls import reverse
 
 from jazzmin.utils import (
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -70,30 +71,12 @@ def test_get_app_admin_urls():
...
-        {
-            "model": "polls.allfields", "name": "Allfields",
-            "url": reverse("admin:polls_allfields_changelist")
-        },
+        {"model": "polls.poll", "name": "Polls", "url": reverse("admin:polls_poll_changelist")},
+        {"model": "polls.choice", "name": "Choices", "url": reverse("admin:polls_choice_changelist")},
+        {"model": "polls.vote", "name": "Votes", "url": reverse("admin:polls_vote_changelist")},
+        {"model": "polls.cheese", "name": "Cheeses", "url": reverse("admin:polls_cheese_changelist")},
+        {"model": "polls.campaign", "name": "Campaigns", "url": reverse("admin:polls_campaign_changelist")},
+        {"model": "polls.allfields", "name": "Allfields", "url": reverse("admin:polls_allfields_changelist")},
```
TysonRV, line 79: Not super opinionated on this, but I think it was easier to read the other way. If we blackenize the whole thing again, this will change again!

tests/test_utils.py
```diff
@@ -108,3 +91,15 @@ def test_get_model_permissions():
...
+@pytest.mark.django_db
+def test_get_model_permissions_lowercased():
+    """
+    When our permissions are upper cased (we had an app with an upper case letter) we still get user perms in lower case
+    """
+
+    user = user_with_permissions("polls.view_poll", "polls.view_choice")
+    user.user_permissions.update(codename=Upper("codename"))
+
+    assert get_view_permissions(user) == {"polls.poll", "polls.choice"}
```
TysonRV, line 105: Nice one 👍

PR #105, opened 2020-08-07
Update icon and add trans tag

fengyouchao:
Change the "change password" icon and add some `trans` tags


farridav:
Nice! Full support for other locales is something we are aiming for

PR #106, opened 2020-08-07
Check fieldsets from admin form, check if fieldset title is None

### LINKED ISSUE #103: When fielsets has only one fieldset, carousel, horizontal_tabs, verical_tabs will not be displayed
When `fielsets` has only one `fieldset`, `carousel`, `horizontal_tabs`, `verical_tabs` will not be displayed

For example:

```
fieldsets = (
    ('Tab1', {
        'fields': (
            'id',
        ),
    }),
)
```

Tab1 will not display.

## Override get_fieldsets

This also affects `get_fieldsets`. Only when `fieldsets` are set and contain more than one `fieldset`, the `fieldsets` returned by the overridden `get_fieldsets` will be displayed correctly.

Case 1

```
# fieldsets is not set
def get_fieldsets(self, request, obj=None):
    return (
        (_('Tab3'), {
            'fields': (
                'id',
            ),
        }),
    )
```

Tab3 will not display.

Case 2:

```
fieldsets = (
    ('Tab1', {
        'fields': (
            'id',
        ),
    }),
)

def get_fieldsets(self, request, obj=None):
    return (
        (_('Tab3'), {
            'fields': (
                'id',
            ),
        }),
    )
```

Tab3 will not display.

Case 3

```
fieldsets = (
    ('Tab1', {
        'fields': (
            'id',
        ),
    }),
    (_('Tab2'), {
        'fields': (
            'release_channel',
        ),
    }),
)

def get_fieldsets(self, request, obj=None):
    return (
        (_('Tab3'), {
            'fields': (
                'id',
            ),
        }),
    )
```

Tab3 will be displayed correctly.

In other words, if `get_fieldsets` is overridden, in order to be displayed correctly, `fieldsets` must be set, and there must be at least two `fieldset`.

farridav:
Fixes #103 

Check more thoroughly if the admin class has no fieldsets

### DIFF
--- a/jazzmin/utils.py
+++ b/jazzmin/utils.py
@@ -204,6 +204,6 @@ def make_menu(user, links, options, allow_appmenus=True):
 
 def has_fieldsets_check(adminform: AdminForm) -> bool:
     fieldsets = adminform.fieldsets
-    if len(fieldsets) == 1 and fieldsets[0][0] is None:
+    if not fieldsets or (len(fieldsets) == 1 and fieldsets[0][0] is None):
         return False
     return True
--- a/jazzmin/utils.py
+++ b/jazzmin/utils.py
@@ -2,7 +2,9 @@ import logging
 from urllib.parse import urlencode
 
 from django.apps import apps
+from django.contrib.admin.helpers import AdminForm
 from django.db.models.base import ModelBase
+
 from jazzmin.compat import NoReverseMatch, reverse
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


jazzmin/utils.py
```diff
@@ -198,3 +200,10 @@ def make_menu(user, links, options, allow_appmenus=True):
             )
 
     return menu
+
+
+def has_fieldsets_check(adminform: AdminForm) -> bool:
+    fieldsets = adminform.fieldsets
+    if len(fieldsets) == 1 and fieldsets[0][0] is None:
```
TysonRV, line 207: mm I guess it can't be zero?

farridav: Nope, djangos get_fieldsets method seems to always return 1 fieldset, unless you override get_fieldsets and return an empty list..

TysonRV: Yeah, it makes sense. I wonder if Django is clever enough to not allow fieldsets to be zero.. probably not 😄 

Otherwise, all looking gooood!

farridav: I've updated it to bool check fieldsets, shoulda woulda coulda written tests

PR #107, opened 2020-08-09
Support for reversing URLs when multiple admins are being used

gkachru:
If multiple admin sites are installed, then Jazzmin's URL reverse code would only point to the default admin. This fixes those URLs thus supporting multiple admin sites.


farridav:
Seems like a reasonable change, I hadn't even realised it was possible to have multiple admin sites.

We don't have any test coverage in this area, So I'm relying on our functional test cases

PR #111, opened 2020-08-13
fix: Remove wrong permission check code

### LINKED ISSUE #110: Side menu does not display models in some cases
When a user does not have `view_user` permission, but overwrites `has_xxxx_permission`  in `UserAdmin`, as follows:

```python
@admin.register(User)
class UserAdmin(admin.ModelAdmin):
    def has_add_permission(self, request):
        return True

    def has_change_permission(self, request, obj=None):
        return True

    def has_view_permission(self, request, obj=None):
        return True

    def has_module_permission(self, request):
        return True
```

Side menu still does not display user.

fengyouchao:
Fixed  #110 

There is no need to check the user's `view_model` permission, because `available_apps` will not include apps without permission, and this permission check code also causes the bug in #110.

### DIFF
--- a/jazzmin/templatetags/jazzmin.py
+++ b/jazzmin/templatetags/jazzmin.py
@@ -25,7 +25,6 @@ from ..utils import (
     order_with_respect_to,
     get_filter_id,
     get_admin_url,
-    get_view_permissions,
     make_menu,
     has_fieldsets_check,
 )
@@ -45,7 +44,6 @@ def get_side_menu(context):
     if not user:
         return []
 
-    model_permissions = get_view_permissions(user)
     options = get_settings()
 
     menu = []
@@ -64,8 +62,6 @@ def get_side_menu(context):
         menu_items = []
         for model in app.get("models", []):
             model_str = "{app_label}.{model}".format(app_label=app_label, model=model["object_name"]).lower()
-            if model_str not in model_permissions:
-                continue
             if model_str in options.get("hide_models", []):
                 continue
 


fengyouchao:
Test failed in `test_no_view_permission`: 

```python
@pytest.mark.django_db
def test_no_view_permission(client):
    """
    When our user has no view permission, they dont see things they are not supposed to
    """
    user = user_with_permissions("polls.change_poll")

    url = reverse("admin:index")
    client.force_login(user)

    response = client.get(url)
    assert parse_sidemenu(response) == {"Global": ["/admin/"]}
```

The above test case seems unreasonable. If the user has `polls.change_poll` permission, this `Poll` should be displayed in the side_menu, because if `jazzmin` is not used, the default admin site will display `Poll`! `ModelAdmin`'s `has_view_permission` by default will return `True` as long as the user has `view_model` or `change_model` permissions.

`has_view_permission` in `BaseModelAdmin`:

```python
class BaseModelAdmin(metaclass=forms.MediaDefiningClass):

    def has_view_permission(self, request, obj=None):
        """
        Return True if the given request has permission to view the given
        Django model instance. The default implementation doesn't examine the
        `obj` parameter.

        If overridden by the user in subclasses, it should return True if the
        given request has permission to view the `obj` model instance. If `obj`
        is None, it should return True if the request has permission to view
        any object of the given type.
        """
        opts = self.opts
        codename_view = get_permission_codename('view', opts)
        codename_change = get_permission_codename('change', opts)
        return (
            request.user.has_perm('%s.%s' % (opts.app_label, codename_view)) or
            request.user.has_perm('%s.%s' % (opts.app_label, codename_change))
        )
```

Therefore, I think the test case `test_no_view_permission` should be removed or modified.

farridav:
I have taken your PR, and added whats needed to it in https://github.com/farridav/django-jazzmin/pull/113

Let me know if this solves the issue, once this is reviewed by @TysonRV i will get a release out to pypi, as we have a few other little fixes to go out too

PR #115, opened 2020-08-16
Allow settings to be over-ridden per-admin and fix for setup_url in a multi-admin setup

gkachru:
_comment/commit removed from pull request, see next comment_


gkachru:
Also added a commit to allow settings override per admin. This allows different admins to have different UI tweaks and settings like varying custom links or search models etc.

If there is an admin with the name 'test' then in the django settings one can add:
`
JAZZMIN_SETTINGS_TEST = {
    'site_title': 'Test'
}

JAZZMIN_UI_TWEAKS_TEST = {
    'navbar': 'navbar-danger navbar-dark'
}
`
which will add/override just those settings in the 'test' admin and not any others.

farridav:
@TysonRV will be taking a look through this PR and thinking about our approach to multiple admins generally, as the documentation will need updating, and we have zero test coverage for multiple admin sites.

Unfortunately this PR no longer fits the title, If you want to address the original bug in a separate PR, and update the title/description of this PR, it will help us to get it through.

Thanks

gkachru:
I've reverted the commit. Will make another PR

gkachru:
Some settings could not be over-ridden (for example in change forms). Have fixed up all the places settings could be called.

farridav:
Closing this until we can come up with a decent way to test and validate multiple admin sites

PR #116, opened 2020-08-17
Remove app.css from password_change_form.html

sepow:
jazzmin/css/app.css was removed at some point, but still exists in the password change form which crashes the app.


TysonRV:
Nice catch! I'll added a new PR with this change here [!117](https://github.com/farridav/django-jazzmin/pull/117), it will be merged in no time.

Thanks @sepow !

PR #119, opened 2020-08-20
Fixing incorrect css/js paths in the JazzminSelect widgets

gkachru:
The css & js paths were incorrect, so the widget didn't change it's look.


farridav:
Good spot, i think this is rarely used / tested, as select2 is automatically used on widgets

PR #128, opened 2020-08-26
Improving the inline admin formset

hwalinga:
There are some issues with the inline admin formset, as can be seen in this picture. 

 * The *-related buttons are overflown to other elements 
 * When the page cannot contain the inline formset, it overflows the box, instead of setting a scroll bar as in the original admin page. 

This PR sets some CSS to resolve this.

Borked admin inline formset


farridav:
Looks good and makes sense

PR #136, opened 2020-09-03
Add custom css js to login and logged_out pages.

hwalinga:
The custom_css and custom_js files should also be added to the login.html and logged_out.html pages as otherwise there is no way to add customized css and js to those pages. 

This PR adds those links in the same way as they are added to the base.html.


farridav:
Nice, easy to review, even easier to merge

PR #138, opened 2020-09-05
fix: Remove title filter

fengyouchao:
The `fieldset.name` should be displayed directly without `title` filter. If  `title` filter is used, it will cause some inappropriate changes. For example, "iOS Opitions" will be changed to "Ios Options".


farridav:
Seems like a reasonable change

PR #139, opened 2020-09-13
Add app index and fix breadcrumb in change list

TysonRV:
This looks like an easy fix but it's becoming a bit tricky... The thing is, the second element (app element) for the breadcrumbs, visible when loading an object for example, is not pointing anywhere. The obvious solution that I found is that the template wasn't in jazzmin, but after adding the template, it was still missing the app list part and adding all apps instead..

Apparently, overriding that one is not that straightforward: https://stackoverflow.com/questions/41770603/how-do-you-override-djangos-admin-app-index

I leave this as food for thought, just in case you have an idea about it!

### DIFF
--- a/jazzmin/templates/admin/app_index.html
+++ b/jazzmin/templates/admin/app_index.html
@@ -1,0 +1,15 @@
+{% extends "admin/index.html" %}
+{% load i18n %}
+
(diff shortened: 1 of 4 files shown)


farridav:
@TysonRV See what you think of the last commit, it might solve your problem .. :)

TysonRV:
@farridav This looks great and it works awesomely! Nice one 🎉

Issue #140, opened 2020-09-15
Group selector height zero


txemavs:
The selector-chosen height is set to 0px in auth/user permissions:
![jazz_group](https://user-images.githubusercontent.com/1844346/93207614-842f5300-f75b-11ea-84fc-a95c6e7cf412.png)

Disabling jazzmin:
![jazz_group_x](https://user-images.githubusercontent.com/1844346/93207809-ce183900-f75b-11ea-9801-bdd8bbe4cb14.png)

 May be JS fails to set selector-chosen height because the tab is not visible on load?

User permissions are ok:
![jazz_user](https://user-images.githubusercontent.com/1844346/93207713-a9bc5c80-f75b-11ea-854d-29f0937755b8.png)

TysonRV:
@txemavs Hey mate! Thanks for reporting. That might be the issue, I'll take a look as soon as I can. 

If you want to try to fix it yourself, feel free to fork the repo and submit a PR!

PR #141, opened 2020-09-15
Fix selector heights on tab shown

txemavs:
Issue [Group selector height zero](https://github.com/farridav/django-jazzmin/issues/140)


TysonRV:
@txemavs I've added some more stuff to it and refactored it a bit on https://github.com/farridav/django-jazzmin/pull/144

It will be on production soon. Thanks for fixing it! 🥇

PR #143, opened 2020-09-17
Jazz up Stacked inlines

TysonRV:
@farridav It took me some time, but I managed to get at least a first version of what I imagine we could get for the stacked inlines. It took me to write some CSS.. but at least no JS (for now 😄 ) 

See screenshots below:

Screenshot 2020-10-03 at 19 05 38
Screenshot 2020-10-03 at 19 06 40
Screenshot 2020-10-03 at 19 06 48
Screenshot 2020-10-03 at 19 06 54

Let me know your thoughts!

### DIFF
--- a/jazzmin/static/jazzmin/css/django.css
+++ b/jazzmin/static/jazzmin/css/django.css
@@ -594,10 +594,8 @@ body.no-sidebar .main-header {
 
 .vCheckboxLabel.inline {
     vertical-align: top;
-    padding-top: 1px;
     color: red;
-    font-style: italic;
-    padding-left: 3px;
+    margin-bottom: 0;
 }
 
 .inline-related .card-header > span {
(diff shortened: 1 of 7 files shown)


farridav:
It's a fantastic start ! I might take a look at some of the UI options available to us, this is currently jazzy AF, and might (date I saw) need toning down a little bit, to ensure that the child content is not more prominent than the parent content.

Exciting shit

TysonRV:
Nice! glad you like it! It definitely needs some more work on it, but I wanted to ship something tangible in the meantime. Probably you have other ideas to enhance them. I thought initially on putting ribbons on the new ones.. but it wasn't very jazzy 😄 

Next I tackle will be the tabulars!

farridav:
Before we get this in, I'll need 5 minutes or so to give some clear recommendations

farridav:
inlines

Ok, so I made some minor aesthetic tweaks, keen to know your thoughts, i have removed some jazz, to make them tidy and not overly prominent ..

TysonRV:
@farridav Yeah, I think it was too prominent indeed. I like this design, it suits the skin a bit better. Just two things:
- I think I introduced a bug, since the remove button should be present in all `new` items
- I would definitely set float right to the `new` items title, to make the difference more obvious. Maybe adding a (+) icon would be nice as well?

TysonRV:
@farridav I'm afraid, the lost `Remove` button from prepopulated fields is something that was happening already.. Probably best to have a new PR for that

TysonRV:
@farridav what about this?

Screenshot 2020-10-04 at 12 29 50

farridav:
That looks much better, you may want to add fa-sm to the add button to keep it the same size as the other icons

PR #152, opened 2020-09-28
Proper support for translations

farridav:
Trying to get translations in a good shape, process is as follows:

1. cd into the jazzmin folder
2. Add the desired language directory e.g mkdir -p locale/de/LC_MESSAGES
3. Run django-admin makemessages
4. Run ./cli.py locales --locale de to remove the django provided strings
5. Go through the strings in the locale file, any that are not genuinely new strings introduced by jazzmin, find
them in the codebase, and try making them match the ones provided in either of djangos translation files
https://raw.githubusercontent.com/django/django/master/django/contrib/admindocs/locale/de/LC_MESSAGES/django.po
https://raw.githubusercontent.com/django/django/master/django/contrib/admin/locale/de/LC_MESSAGES/django.po

Once finished, run makemessages again, until the file contains ONLY unique strings to jazzmin, there should
only be a handful

6. Translate those strings


tests/test_app/settings.py
```diff
@@ -78,9 +78,17 @@
...
 USE_I18N = True
+LOCALE_PATHS = (os.path.join(BASE_DIR, "locale"),)
 USE_L10N = True
 USE_TZ = True
 
+# English default
+gettext = lambda s: s  # NOQA
+LANGUAGES = (
+    ("en", gettext("English")),
+    ("de", gettext("German")),
```
TysonRV, line 89: I'll add Spanish 😄

TysonRV:
@farridav this is great dude! Really good job!

tests/test_app/settings.py
farridav: You can help with this if you want, i need to whittle down all the strings in this translation file until its just jazzmin ones, there should be about 5 strings left

PR #153, opened 2020-09-28
Spanish language 

TysonRV:
- Add Spanish as a language
- Modify some strings to make more sense on different languages
- Bits on German to adapt to new changes


TysonRV:
@farridav The only blocker that I saw by generating the language was the fact that I didn't have `polib` installed. Since this is something quite specific to people who wants to translate, I don't think it makes sense to add it to Poetry. What it might make sense is to update the docs with the info that resides on `cli.py` at the moment, what do you think?

farridav:
Amazing, yes I didn't add polib to the deps, though we could add it to dev deps if need be, the process needs a bit more work, as it's a bit unorthodox having to remove translation strings in that way.

A couple of things,

1. There are fuzzy strings in the translations (these will not be carried into compiled translation files)

2. We are still translating some strings that Django has already translated into many other languages

The reason for no.2 is because we have formatted the string slightly differently, so the msgid doesn't match up, we can fix this by looking at how it's defined in djangos .po files and adjusting our strings.

Getting rid of fuzzy tags is as easy as removing it from mo files, we might want to use something like Rosetta for administering these (it probably has polib as a sub dependency)

farridav (approved):

PR #154, opened 2020-09-29
refactor: Use subparser

fengyouchao:
A more elegant way to implement subcommands


farridav (approved):
Nice, let's keep chipping away at the process and smoothing it out

PR #165, opened 2020-10-06
Add has add permission to fix bug on tabular inlines

### LINKED ISSUE #125: TabularInline does not display the last data and missing "add another {model}" link
## 1. When a `TabularInline` does not have add permission, the last inline data will not be displayed.

Example:

```python
class AnswerInline(admin.TabularInline):
    extra = 0
    model = Answer

    def has_add_permission(self, request, obj=None):
        return False


@admin.register(Question)
class QuestionAdmin(admin.ModelAdmin):
    inlines = (AnswerInline,)
```
If a `question` has two `answers`, but only the first answer will be displayed in `QuestionAdmin`.

## 2. Missing "add another {model}" link

`TabularInline` does not display the "add another {model}" link even if it has add permission.

TysonRV:
Fixes #125

### DIFF
--- a/jazzmin/templates/admin/edit_inline/tabular.html
+++ b/jazzmin/templates/admin/edit_inline/tabular.html
@@ -31,7 +31,7 @@
                         </td>
                     </tr>
                     {% endif %}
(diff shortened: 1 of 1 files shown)


TysonRV:
@farridav I'll close this on favour of #167

Issue #169, opened 2020-10-07
Compatibility with django-mptt


dimastbk:
Hi!

Can you replace `<table class="table table-striped">` with `<table id="result_list" class="table table-striped">` in `change_list_results.html`? (This "id" is in the original django admin.) This want for compability with `DraggableMPTTAdmin` from [django-mptt](https://github.com/django-mptt/django-mptt).

Also, this style impove view DraggableMPTTAdmin in jazzmin
```
#result_list .field-tree_actions {
    width: calc(40px + 2.25rem);
}

#result_list .field-tree_actions > div {
    margin-top: 0;
}
```

Thanks.

dimastbk:
Also, `TreeRelatedFieldListFilter` is broken in jazzmin. But i don't know, how well fixed it.

farridav:
Sure this seems quite reasonable, ill take a look today

farridav:
I've been looking through this, and its not as simple as i had hoped, django-mptt seem to be using their own templates (see https://github.com/django-mptt/django-mptt/tree/master/mptt/templates/admin)

and regardless of where in `INSTALLED_APPS` i put `mptt` I never get the right template, my WIP branch is here https://github.com/farridav/django-jazzmin/pull/170

Im trying to see if i can affect the template, but as a last resort, we can add necessary styling through CSS, or use JS

So far, the page doesnt look as expected, though i may not have MPTT configured properly, happy to receive guidance on this..

farridav:
Ok, i have it all rendering out nicely now, thanks .. i just need to get the tests updated, and we will get it in ..

Issue #126, opened 2020-08-24
Compatability with django-treebeard


EhsanSafir:
when add  django-treebeard library to django project admin not show add button for tree models

>
![Screenshot from 2020-08-24 16-53-56](https://user-images.githubusercontent.com/58684770/91044634-d8e32080-e62a-11ea-9eda-d73838a1d827.png)

farridav:
Resolved

PR #170, opened 2020-10-07
Provide support for mptt

farridav:
- Allow django-mptt templates to render out as expected

See https://github.com/django-mptt/django-mptt/tree/master/mptt/templates/admin

Fixes: https://github.com/farridav/django-jazzmin/issues/169 and https://github.com/farridav/django-jazzmin/issues/126

Note: selects (dropdowns) not using select2 on these templates, as it is not rendering our change form, this not bringing in select2.min.js, i might need to work on this


dimastbk:
For treefilter can use this:

change_list.js
```js
$(document).ready(function () {
    // Ensure all raw_id_fields have the search icon in them
    $('.related-lookup').append('<i class="fa fa-search"></i>');

    // Allow for styling of selects
    $('.actions select').addClass('form-control');

    // Make search filters select2 and ensure they work for filtering
    let $ele = $('.search-filter');
    $ele.search_filters();
    $ele.select2();

    let $mptt = $('.search-filter-mptt');
    $mptt.search_filters();
    $mptt.select2({
        dropdownAutoWidth: true,
        templateResult: function (data) {
            // https://stackoverflow.com/questions/30820215/selectable-optgroups-in-select2#30948247
            // rewrite templateresult for build tree hierarchy
            if (!data.element) {
                return data.text;
            }
            let $element = $(data.element);
            let $wrapper = $('<span></span>');
            $wrapper.attr('style', $($element[0]).attr('style'));
            $wrapper.text(data.text);
            return $wrapper;
        },
    });
});
```

mptt_filter.html
```html
{% load i18n %}
{% load mptt_admin %}

<div class="form-group">
    <select class="form-control search-filter-mptt" style="width: 100%;" tabindex="-1" aria-hidden="true" data-name="{{ field_name }}">
        <option value="">{{ title }}</option>
        <option value="">---------</option>
        {% for choice in choices %}
            {% if choice.name %}
                <option {{ choice.padding_style }} data-name="{{ choice.name }}" value="{{ choice.value }}" {% if choice.selected %}selected {% endif %}>
                    {{ choice.display }}
                </option>
            {% endif %}
        {% endfor %}
    </select>
</div>
```

Result:
изображение

TysonRV:
@farridav I think this works fine. I would need to play around with mptt a bit more though. The only bug I've managed to get, that I'm sure it is coming from jazzmin, is that ..when there is no objects yet, the add button disappears. This doesn't happen when there is objects tho!
I'm using it on my project, so I'll keep an eye on it. So far, it looks like it is all functional 👍

TysonRV:
Oh, just saw @dimastbk comments. I guess we can open up a new one for those!

@dimastbk Feel free to send a PR to the repo! Here some guidelines: https://django-jazzmin.readthedocs.io/development/

Thanks for flagging the issue in the first place 🥇

PR #171, opened 2020-10-09
Fix translations

fengyouchao:
Corrected the wrong translation and changed some wording in zh_Hans and zh_Hant


TysonRV:
@fengyouchao thanks for this!

PR #173, opened 2020-10-12
(fix) capfirst for label

dimastbk:
If label of models field in lowercase and this field set as readonly in admin, then this label was lowercase in admin.


TysonRV:
@dimastbk Awesome, nice catch!

PR #175, opened 2020-10-12
Bits and bobs are back

### LINKED ISSUE #97: Lock down demo user/app more, avoid trolling
It seems someone had changed the password for the demo, stopping other people from logging in, 

Make the reset command more belt and braces, and tighten up the admin user (dont use superuser)

TysonRV:
- Fix bug on m2m filters
- Redo poetry lock
- Organise models and fixtures
- Locks test user (add only `test_group` permissions) fixes #97


pyproject.toml
```diff
@@ -42,6 +42,7 @@ exclude = ["tests", "docs"]
 [tool.poetry.dependencies]
 python = ">=3.5"
 django = ">=2"
+dj_database_url = "^0.5.0"
```
farridav, line 45: This is not a dependency for jazzmin, dependencies for the test app are declared in tests/test_app/requirements.txt

farridav (changes requested):

pyproject.toml
TysonRV: Oh, that's so right..

TysonRV: @farridav the problem is that, according to the docs, the only thing a new dev has to do is to do `poetry install`. This won't work 100%, as we import `dj_database` module right on the settings.. But obviously for development you don't need all the packages that are on `requirements.txt`

farridav (approved):

PR #176, opened 2020-10-12
Django import export support

TysonRV:
- Add templates for jazzing [`django-import-export` package](https://django-import-export.readthedocs.io/en/latest/)


farridav:
Wow, there is a lot of stuff here, I'll need to really dig into import export to see why so much is needed.

Once this is code complete, we will also need to update the docs

TysonRV:
> Wow, there is a lot of stuff here, I'll need to really dig into import export to see why so much is needed.
> 
> Once this is code complete, we will also need to update the docs

Yeah! just wanted to know your opinion on this really. I think this is the most flexible, although it could be reduced for sure. Not really 100% whether we wanted to add it tho

farridav:
I think supporting it is a great idea, I'll check it out locally asap and try to understand the changes being made

farridav:
Let's get this debased, and demonstrate its usage in the test app

TysonRV:
@farridav I've added usage for Books, so we have this on the test_app. Some styling still needed, but getting in shape:

Screenshot 2020-10-15 at 13 00 21

Screenshot 2020-10-15 at 13 00 11

Screenshot 2020-10-15 at 13 00 28

PR #177, opened 2020-10-12
Format submit line and add block for content

TysonRV:
- I've added `{% block submit-row-content %}` to `submit_line.html` to add a bit more of flexibility to it, as right now to extend that template it would be necessary to add most of the code (or new buttons will be added outside the card)

Inspired on https://code.djangoproject.com/ticket/28267

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -297,3 +297,43 @@ custom CSS and/or JS file, just pass a relative path to your files e.g:
  ```
 
 Into your jazzmin settings (Ensure these files can be found by the static file finder)
+
+### Adding extra actions to model's form view
+
+Add a template for your model on your main template directory
+
+```djangourlpath
+your_app/templates/admin/your_model/change_form.html
+```
+
+```jinja2
+{% extends "admin/submit_line.html %}
+
+{% block extra-actions %}
+
+{# For a simple link #}
+<div class="form-group">
+    <a href="/some/url" class="btn btn-outline-info form-control">Some Action</a>
+</div>
+
+{# Or, to process this with form submission #}
+<div class="form-group">
+    <input type="submit" class="btn btn-outline-info form-control" value="SomeAction" name="_your_action">
+</div>
+{% endblock %}
+```
+
+If you are adding a button that needs processing with the form, e.g (Save and send) you will need to add the
+following to your admin class:
+
+```python
+def response_change(self, request, obj):
+    ret = super().response_change(request, obj)
+    if '_your_action' in request.POST:
+        # Do something
+        pass
+    return ret
+```
+    
+The implementation might change slightly if your wanting to perform an action on add, or delete, for those, you can
+override the response_add of response_delete methods instead/as well.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -303,11 +303,11 @@ Into your jazzmin settings (Ensure these files can be found by the static file f
 Add a template for your model on your main template directory
 
 ```djangourlpath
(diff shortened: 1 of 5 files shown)


farridav:
If someone overrides `submit-row-content` they still need to copy-pasta quite a lot of code right ? should we localise this tag a bit more ? its incredibly useful to have :) and i think we should add it to the docs too !

farridav (commented):
Perhaps:

```html
{% load i18n admin_urls %}

{% block submit-row %}
<div class="card card-primary card-outline">
    <div class="card-header">
        <h3 class="card-title">
            <i class="fas fa-edit"></i>
            {% trans 'Actions' %}
        </h3>
    </div>
    <div class="card-body">
        {% if show_save %}
        <div class="form-group">
            <input type="submit" value="{% trans 'Save' %}" class="btn btn-success form-control" name="_save">
        </div>
        {% endif %}
        {% if show_delete_link and original %}
            <div class="form-group">
                {% url opts|admin_urlname:'delete' original.pk|admin_urlquote as delete_url %}
                <a href="{% add_preserved_filters delete_url %}" class="btn btn-danger form-control">{% trans "Delete" %}</a>
            </div>
        {% endif %}
        {% if show_save_as_new %}
            <div class="form-group">
                <input type="submit" class="btn btn-outline-info form-control" value="{% trans 'Save as new' %}" name="_saveasnew">
            </div>
        {% endif %}
        {% if show_save_and_add_another %}
            <div class="form-group">
                <input type="submit" class="btn btn-outline-info form-control" value="{% trans 'Save and add another' %}" name="_addanother">
            </div>
        {% endif %}
        {% if show_save_and_continue %}
            <div class="form-group">
                <input type="submit" class="btn btn-outline-info form-control" value="{% if can_change %}{% trans 'Save and continue editing' %}{% else %}{% trans 'Save and view' %}{% endif %}" name="_continue">
            </div>
        {% endif %}
        {% if show_close %}
            <div class="form-group">
                <a href="{% url opts|admin_urlname:'changelist' %}" class="btn btn-outline-danger form-control">{% trans 'Close' %}</a>
            </div>
        {% endif %}
        {% block extra-actions %}
        {% endblock %}
    </div>
</div>
{% endblock %}
```

And in the docs, we can add a section stating how you add extra-actions e.g:

Add a template for your model

```html
# your_app/templates/admin/your_model/change_form.html
{% extends "admin/change_form.html %}

{% block extra-actions %}

{# For a simple link #}
<div class="form-group">
    <a href="#SomeUrl" class="btn btn-outline-info form-control">Some Action</a>
</div>

{# Or, to process this with form submission #}
<div class="form-group">
    <input type="submit" class="btn btn-outline-info form-control" value="SomeAction" name="_your_action">
</div>
{% endblock %}
```

If you are adding a button that needs processing with the form, e.g (Save and send) you will need to add the 
following to your admin class:

```python

def response_change(self, request, obj):
    ret = super().response_change(request, obj)
    if '_your_action' in request.POST:
        # Do something
        pass
    return ret
```

The implementation might change slightly if your wanting to perform an action on add, or delete, for those, you can 
override the `response_add` of `response_delete` methods instead/as well.

TysonRV:
@farridav Thanks for the suggestions, definitely taking that approach! 👍 

Won't wrap the `save` button tho, as that would limit the functionality of the form, i.e, you want to keep the `close` button but not have the `save` one.

TysonRV:
@farridav I also added a practical example to the test app

farridav (approved):

PR #185, opened 2020-10-16
Fix minor errors on test app

TysonRV:
- At the moment, it's not possible to save a book due to some required fields missing
- Made duration a read field


farridav:
LGTM

PR #187, opened 2020-10-17
Related modal bootstrap

KarimTayie:
First of all thanks for the job done on this amazing project.
Second, in this PR I have added a flag in jazzmin settings to render Django popups inside an iframe within bootstrap 4 modals. (instead of the old popup window, optional)
This PR is heavily inspired by https://github.com/fabiocaccamo/django-admin-interface related model feature.

Before:
Screenshot from 2020-10-17 20-44-45

After:
Screenshot from 2020-10-17 19-24-38

Feel free to leave any comments.


farridav:
I really like the idea of this, and thank you for your contribution!

I'm going to try and look over this during the week, mostly to ensure it has suitable levels of:

- maintainability
- jazzyness
- testing if necessary

jazzmin/static/jazzmin/js/related-modal.js
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            relatedModalCounter+=1;
+
+            $.showModal({
+                title: title,
+                body: body,
+                modalDialogClass: "modal-dialog-centered modal-lg",
+                modalClass: "related-modal-" + relatedModalCounter,
+            });
```
farridav, line 39:
```javascript
            $.showModal({
                title: title,
                body: body,
                backdrop: false,
                modalDialogClass: "modal-dialog-centered modal-lg",
                modalClass: "fade modal-wide related-modal-" + relatedModalCounter,
            });
```

Adds `fade`, and `modal-wide` into modalClass

farridav (changes requested):
1. When an item is added via an n+1 popup it renders out a blank page and does not close, see 
image
2. The escape key can only close the latest popup, any underneath it are unclosable using escape
3. the dynamic modal need a couple of classes added to it (comment inline)

That's all I've found so far, generally the user experience is quite good, if we can address the above its good to go in.

KarimTayie:
@farridav 
I have pushed new commits for the 3 above comments.

farridav:
Fantastic work! everything behaves nicely, this is a great feature

PR #188, opened 2020-10-20
Related modal bootstrap

farridav:
Updates https://github.com/farridav/django-jazzmin/pull/187

Adds/Changes the following:

- Use const/let in JS instead of var
- No need for explicit < Django 2 support, as jazzmin doesnt support it
- Improve rendering of page in popup mode (remove breadcrumbs)
- Increase size of modal
- populate modal title from link title
- Blur modal background
- Update docs
- re-do screenshot to reflect changes
- blacken codebase

Fixes https://github.com/farridav/django-jazzmin/pull/187

@KarimTayie let me know if this is in-line with your expectations, if it is, we will get it in ASAP, its a great enhancement, props to https://github.com/fabiocaccamo/django-admin-interface too, for the original implementation


KarimTayie:
@farridav 
Everything looks good to me but after some manual testing, I found this case when u try to open multiple modals inside each other they look a bit weird.
Peek 2020-10-21 03-38
So after a few thoughts, I decided to use a more dynamic approach with bootstrap modal. and here is how it looks when opening multiple modals inside each other
Peek 2020-10-21 03-30
I think it looks better, but it may need some testing. so if u wanna merge the current one as it is, it is ok with me. I kept all your changes in this new edit.

farridav:
This change makes a lot of sense, I'll close out this PR, and will need to interview the changes again.

PR #199, opened 2020-11-03
Define a max width for the login form

### LINKED ISSUE #198: Max width in login form with error
I would like to suggest to define a `max-width: ~700px` for the login form, or at least when there are errors.

marcorichetta:
Fixes #198

I found this to be the simplest way. Let me know if there are other things to take into account.

Also, I can't find where `register-box` is being used. Maybe it's left over from before?


farridav:
LGTM there are some unused css classes yes, I'll clean them up at some point :)

PR #200, opened 2020-11-03
Improve style of related modal

dimastbk:
Some improve style for #187 
- style as native dialog
- set title dialog from page title from iframe (for nested dialogs also)
- removed the hard code `width: 50%`, because it broke dialog in the mobile view
изображение
изображение

There are some bugs in this PR...

### DIFF
--- a/jazzmin/static/jazzmin/css/django.css
+++ b/jazzmin/static/jazzmin/css/django.css
@@ -721,21 +721,6 @@ a.inline-deletelink {
     margin-top: 0;
 }
 
-/* wide modals */
-.modal.modal-wide .modal-dialog {
-    width: 50%;
-    max-width: inherit;
-}
-
-.modal-wide .modal-body {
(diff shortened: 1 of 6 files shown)


TysonRV:
@dimastbk Thanks for the contribution! 

I need to have a closer look to this, but will review it shortly. Also, @farridav I think it makes sense to test this on your widescreen as well!

farridav:
I'm afraid were not quite here with this

Desktop:
Old: 
image

New:
image

The scrollbar is appearing on the inner content when not needed, the modal is not fitting the content

I can see that on mobile, all popups are in a pretty bad state, something we obviously didnt check properly in the first wave of changes.. I suggest we default popup in modal to off until we have this working as expected on all target screen sizes

farridav:
I've defaulted the use of morals to off for now (people can still opt in)

We should address the display issues on mobile either in this PR, or separately

PR #201, opened 2020-11-04
Support for hu locale

kristofgilicze:
- added translations for hungarian.
- removed the extra strings with the cli tool ( as requested in dev docs )
- added **'hu'** locale to options


TysonRV:
Hi @kristofgilicze ! 

Thanks for the PR, it's great to be able to jazz up the admin in Hungarian too! 

Note that some of the strings that are on that `po` file are coming directly from Django. You could just run `./cli.py locales --prune hu` to remove all of them and make it easier to maintain in the future.

I'll merge the PR once that's done 😄

kristofgilicze:
Hello !

I did ran the tool, it removed at least 400 lines from `django.po`.
So either the tool is broken or my env ?


If all else fails, then I can manually compare with the other languages.

TysonRV:
@kristofgilicze Hey!
You are actually right. The problem is that we have started supporting now other apps, e.g, `django-import-export`, and there are some translations coming from those which we didn't count on initially.

This will require some further investigation, so I will merge this PR already to keep things moving.

Thanks again 🙂

PR #204, opened 2020-11-06
Allow camel cased django app names. Fixes #203.

### LINKED ISSUE #203: utils.get_admin_url goes wrong for apps named with camel case
I think there may be a problem with [get_admin_url](https://github.com/farridav/django-jazzmin/blob/2b51e6ee1efeaea415376a4fe6b1175d2362e1ad/jazzmin/utils.py#L49) and apps named in CamelCase.

Given we have an app named `CamelStore` and inside a model named `CamelType` the admin change list name for that would be `admin:CamelStore_cameltype_changelist`. As `get_admin_url`  converts the whole string with `lower()` that would result in a non resolvable admin url.

Just removing the `lower()`call in  [line 50](https://github.com/farridav/django-jazzmin/blob/2b51e6ee1efeaea415376a4fe6b1175d2362e1ad/jazzmin/utils.py#L50) should solve that problem.

_Originally posted by @mpibpc-mroose in https://github.com/farridav/django-jazzmin/issues/99#issuecomment-721690070_

mpibpc-mroose:
Model admin URLs for camel cased app names where not resolved correctly. This was solved by removing the use of lower() on the app name.

### DIFF
--- a/jazzmin/settings.py
+++ b/jazzmin/settings.py
@@ -132,6 +132,17 @@ CHANGEFORM_TEMPLATES = {
 }
 
 
+def get_search_model_string(jazzmin_settings):
+    """
+    Get a search model string for reversing an admin url.
+
+    Ensure the model name is lower cased but remain the app name untouched.
+    """
+
+    app, model_name = jazzmin_settings["search_model"].split(".")
+    return "{app}.{model_name}".format(app=app, model_name=model_name.lower())
+
+
 def get_settings() -> Dict:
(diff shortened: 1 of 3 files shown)


farridav:
This looks good :) Happy to get it in if you are @TysonRV

TysonRV:
This looks good to me! It is missing typing for the new method `get_search_model_string`, but everything else looks great.

Thanks for the contrib! @mpibpc-mroose

TysonRV (approved):

PR #206, opened 2020-11-07
Force test@test.com user to use test password

### LINKED ISSUE #205: Don't let the superuser change its password in the demo admin page
Right now the superuser test@test.com can change its password in demo page. Don't give the permission to that user to do so.

farridav:
When updating the superuser, force the use of our known password, whilst preserving the ability to view and interact with all parts of the admin 

Fixes #205

### DIFF
--- a/tests/test_app/library/books/receivers.py
+++ b/tests/test_app/library/books/receivers.py
@@ -1,0 +1,12 @@
+from django.contrib.auth.models import User
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
+
+
+@receiver(pre_save, sender=User)
+def add_user_levels_actions(sender, instance, **kwargs):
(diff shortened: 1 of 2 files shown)


tests/test_app/library/books/receivers.py
```diff
@@ -0,0 +1,12 @@
+from django.contrib.auth.models import User
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
+
+
+@receiver(pre_save, sender=User)
```
TysonRV, line 6: Shouldn't it be in a `post_save` signal instead?

farridav: nope, this is setting the password in on the instance before it is saved, using post_save would cause a double save

PR #207, opened 2020-11-07
Theming support via bootswatch

farridav:
- Add in theming support via UI customiser
- Move all library static files into vendor
- Adds in dropdown for choosing theme in UI customiser drawer
- Adds in all current bootswatch themes
- Adds `theme-{theme_name}` into body classes (for theme specific CSS rules)
- Disables modal popups for now by default (can be enabled), as they are broken on mobile
- cleans up and groups sidebar customisations
- Adds the following new UI tweaks:
  - Whether to constrain the page to a box (leaving big margins at the side) with `"layout_boxed": False,`
  - Make the footer sticky, keeping it in view all the time with `"footer_fixed": False,`
  - Make the sidebar sticky, keeping it in view as you scroll with `"sidebar_fixed": False,`
  - Make the top navbar sticky, keeping it in view as you scroll with `"navbar_fixed": False,`


jazzmin/static/jazzmin/css/django.css
```diff
@@ -105,6 +105,10 @@ table.dataTable thead .sorting_desc_disabled::after {
     height: 40px !important;
 }
 
+.select2-results__option {
```
farridav, line 108: really not sure why i needed this, but the option text is white without it

jazzmin/static/jazzmin/js/ui-builder.js
```diff
@@ -140,9 +140,23 @@
...
         });
+
+        $("#jazzmin-theme-chooser").on('change', function () {
+            const $themeCSS = $('#jazzmin-theme');
+            const currentSrc = $themeCSS.attr('href');
+            const currentTheme = currentSrc.split('/')[4];
+            const newTheme = $(this).val();
+            $themeCSS.attr('href', currentSrc.replace(currentTheme, newTheme));
+
+            if (newTheme === "darkly") {
```
farridav, line 151: If certain UI tweaks are recommended with a theme, I'm selecting them here, e.g a dark navbar

kristofgilicze:
The scenario that I can't solve this way is:
Let's say I want `flatly` normally and `darkly` for the dark pref. 

**I propose something like this:**
( I feel like this is a bit easier to understand, needs less explanation. It is also entirely possible that I have tunnel vision on this one )

```python
# `flatly` is default, `darkly` when user has 'dark mode pref'.
UI_TWEAKS {
  ...
  'theme': 'flatly',
  'dark_theme': 'darkly'
  ...
}

# `bundled boostrap` is default, `darkly` when user has 'dark mode pref'.
UI_TWEAKS {
  ...
  'theme': None,
  'dark_theme': 'darkly'
  ...
}

# `flatly` is default, darkmode is disabled.
UI_TWEAKS {
  ...
  'theme': flatly,
  'dark_theme': None
  ...
}

# same as before
UI_TWEAKS {
  ...
  'theme': None,
  'dark_theme': None
  ...
}

```
I don't really see the benefit of setting the `theme_condition` in the config.

Template should be something like this with the setup above:
>```django
>  
>  {% if  jazzmin_ui.theme %}
>    <link rel="stylesheet" href="{{ jazzmin_ui.theme }}"  id="jazzmin-theme" />
>  {% endif %} 
>  {% if  jazzmin_ui.dark_theme %}
>    <link rel="stylesheet" href="{{ jazzmin_ui.dark_theme }}" media="(prefers-color-scheme:dark)"/>
>  {% endif %} 
> ```

farridav:
The scenario of having two themes (apart from the default one) is a new requirement, I do think you may have tunnel vision on this one a bit, though that doesn't mean to say your not giving valuable suggestions.

What I have tried to achieve here, and my primary goal was to provide theming support with bootstwatch, and in doing so, allow the use of a dark theme, namely darkly.

The second requirement I have tried to fulfill is allowing the users system to determine which theme to use, and since this setting is currently limited to light/dark, I had achieved this using the default template (which is light) or a developer chosen dark template, with a pre condition.

It's not that I've misunderstood what your asking for, it's that we started with two different goals (dark mode, and a developer chosen theme), and I'm trying to accommodate for both.

The latest suggestion of having a theme, and a dark theme could work, and is the most flexible permutation so far, the only challenge is to explain (in the docs) that if you want to use a dark theme regardless of system preference, you need to set it in "theme" and leave "dark_theme" as None, which is a bit counter intuitive, especially to those that are new to the web standard.

farridav:
@kristofgilicze I have pushed up some more changes to enable an different theme for dark mode and (light mode or no preference)

I think this combines all of the different directions together, and hopefully isnt too complicated for configuration (simple, easy, optional configuration is one of jazzmins USP's)

Let me know if this brings us all into alignment, be sure to check out the implementation, the documentation, and the UI customiser

:) Happy friday, lets aim to get this shipped today !

kristofgilicze (approved):
**I have tested it, looks good to me!**

> The scenario of having two themes (apart from the default one) is a new requirement

Yes, I understand the original goal was to have 'Theming support via bootswatch'.


> The latest suggestion of having a theme, and a dark theme could work, and is the most 
flexible permutation so far, the only challenge is to explain (in the docs) that if you want to use a dark theme regardless of system preference, you need to set it in "theme" and leave "dark_theme" as None, which is a bit counter intuitive, especially to those that are new to the web standard.

I think this is fine hidden away, if someone is interested then the docs explain how does it work under the hood.

farridav:
Awesome sauce, @TysonRV can I get your eyes on this today ? I know it's a big one, but it's be awesome to ship this today if possible.. it being Friday and all :)

jazzmin/static/jazzmin/js/ui-builder.js
TysonRV: wouldn't this be recommended to all dark themes then?

tests/test_app/library/settings.py
```diff
@@ -200,7 +200,7 @@
     # Related Modal #
     #################
     # Activate Bootstrap modal
-    "related_modal_active": True,
+    "related_modal_active": False,
```
TysonRV, line 203: Why did you deactivate this for the test app?

farridav: I've deactivated it everywhere until we solve the mobile issue (modals are not working properly on mobile)

People can still opt to use them if they want, as it's pretty decent on desktop, hopefully we will address the issue in that other PR

TysonRV (approved):
Everything looking smashing here!

Issue #209, opened 2020-11-07
ForeignKey select not populated when adding new inline object


Blaza:
Hi! I love the package and the style of the admin, so thanks for creating it.

There seems to be a bug in inline form handling with foreign keys, that's even present on the demo app.

For example, if you go to https://django-jazzmin.herokuapp.com/en/admin/books/author/44/change/#books-tab
and click "Add another book", the "Library" select box for the new row will have no items available, whereas in the initial inline rows you can choose libraries from the select box. 

I attach a couple of screens:
- a preexisting row (because of the `extra` parameter probably) which has options available as expected
![regular](https://user-images.githubusercontent.com/693523/98449108-36b6dd00-2131-11eb-98e9-593102e29cef.png)

- and of the newly added row, for which there are no choices populated
![newly_added](https://user-images.githubusercontent.com/693523/98449110-374f7380-2131-11eb-99bb-99a361da1821.png)

Hopefully something can be done here.

I tested it on Ubuntu on Chrome 86 and FIrefox 81 and it's the same.

farridav:
Hmm, sounds odd, does the problem exist in normal Django admin if you comment jazzmin out from your installed apps ?

To summarise, the problem is that new inline records that have a foreign key selector, appear to have no available choices?

I'll check out the demo app shortly.

Thanks for reporting, and glad your enjoying the package

Blaza:
Nope, the default admin works as expected. I'm on Django 3.0 btw.

frnmst:
Hi,

this happens with Django 3.1 as well.

Blaza:
Ok I managed to fix it by
1. Changing the line https://github.com/farridav/django-jazzmin/blob/49241820d20afd6f6cb7312aaa80aecb521a51b0/jazzmin/static/jazzmin/js/change_form.js#L123

to 
```
    const noSelect2 = '.empty-form select, .select2-hidden-accessible, .selectfilter, .selector-available select, .selector-chosen select';
```
Note the `.empty-form select`. Django makes an empty-form for inlines, and this line was making it a select2, which screwed up the ids. Every new row would have something like `data-select2-id="id_book_set-__prefix__-library"` i.e. not changing the `__prefix__` part because it copied it from empty form

2. Making all selects be select2 upon adding a formset by adding the following at the end of `jazzmin/js/change_form.js`
```
django.jQuery(document).on('formset:added', function(e) {
[...]

PR #211, opened 2020-11-08
Fix select2 for newly added inline rows

Blaza:
Solves issue #209, where new inline records that have a foreign key selector appear to have no available choices.


farridav:
Nice work, clean, concise and accurate

PR #212, opened 2020-11-09
Replaces "{ }" with escaped braces "{{ }}" fix for json format string error

carderm:
A json field with curly braces can cause an error inside the django function 'format_html' with 'key error'. This resolves that.


farridav:
I'd like to be able to test this, also perhaps the problem goes away if we change format_html to mark_safe ..

carderm:
> I'd like to be able to test this, also perhaps the problem goes away if we change format_html to mark_safe ..

It seems to work with mark_safe - might be a cleaner solution. I'll commit that.

TysonRV:
@carderm Looking good! Thanks for contributing 🙂

PR #213, opened 2020-11-09
Added typing and test for get_search_model_string.

mpibpc-mroose:
As mentioned in PR for #203 and #204 tests and typing should get added for `get_search_model_string`


TysonRV:
@mpibpc-mroose Thanks for this! 

It looks good to me. What do you think @farridav ?

farridav:
Looking good :)

PR #217, opened 2020-11-17
Escape messages before marking them as safe

sacovo:
the style_bold_first_word template filter marks the message strings as safe, this causes an issue if any words in the message contain critical chars, for example if the str of an object contains `<` or `>`, the layout is broken if such objects are in the message.


farridav:
Seems reasonable

PR #218, opened 2020-11-19
Fix broken translation

ethemce:
translation of "Yes, I’m sure" not working


TysonRV:
@ethemce This makes sense to me. Thanks for the contribution!

Issue #220, opened 2020-11-29
Support for django-admin-charts


yswtrue:
When I add django-admin-charts, get this error, when I change jquery-slim to jquery the problem fixed.
```
jquery.js:4055 Uncaught TypeError: $.ajax is not a function
    at loadChart (admin_charts.js:21)
    at HTMLFormElement.loadAnchor (admin_charts.js:45)
    at Function.each (jquery-3.5.1.slim.min.js:2)
    at E.fn.init.each (jquery-3.5.1.slim.min.js:2)
    at HTMLDocument.<anonymous> (admin_charts.js:52)
    at mightThrow (jquery.js:3762)
    at process (jquery.js:3830)
```

farridav:
Curious, I would have thought Django admin charts would ship with its own copy of jQuery used with noConflict.

I swapped ours out for the slim version a little while back,Nas we don't use the full library, and it makes for faster page load.

Perhaps we can remove our copy of jQuery entirely, as the one from Django is new enough to support the plugins we have in use (AFAIK)

Or we can try using noConflict on our jQuery slim lib, which will relinquish the $ and allow Django charts to whatever jQuery is making use of the $ variable

Issue #219, opened 2020-11-22
Some elements need additional styling when a selected theme is a dark one. 


hwalinga:
A continuation of #192 

Some elements aren't proper bootstrap elements and need manual styling when the selected theme is a dark one. 

These elements are:

* The login page: 
  - .login-page, .register-page
  - .login-card-body, .register-card-body
* The "timeline":
  - .timeline > div > .timeline-item
  - .timeline > div > .timeline-item > .time

The last one requires a `color` to be set for the text color, the others just a `background`. 

The header: A simple and clean work-around for this is set `"navbar": "navbar-dark bg-primary"` in the `JAZZMIN_UI_TWEAKS`.

You probably can solve this with just loading like four lines of additional CSS when a dark theme is selected.

farridav:
I was hoping there would be some follow up tweaks for dark mode, and it seems like you have a fairly clear idea of some things that need doing.

Perhaps you can make a PR  instead ?

hwalinga:
I could, but my sense for colors is horrible and I am not a dark theme user. 

But I can make a draft and then you can change the colors to something more pleasing to the dark theme users.

farridav:
Perhaps there are bootstrap classes or changes in the markup that we could make to these elements to help them fall in line with the theme ?

I'll take a look at this too when I can .

hwalinga:
Hmmm... Well I am not very familiar with Bootstrap either...

But let's see what I make of it. I just thought of loading in some custom CSS when the theme is a dark one.

farridav:
Probably no need for a new static file, as we can target themes using CSS selectors, I'll take a look too, I'm quite familiar with bootstrap and adminlte.

Anything that looks out of place, is usually something we can remedy by using the right markup or bootstrap classes .

I'll also take a look at this early next week

PR #221, opened 2020-11-30
Dark theme tweaks, button customisation

farridav:
- Allow changing of button types (outline/solid)
- Add in all languages to test app supported by Django
- fixup login page and timeline boxes so they look better in dark themes
- Add default UI tweaks into test app
- uses djangos copy of jQuery instead of slim (should fix https://github.com/farridav/django-jazzmin/issues/220)

Fixes https://github.com/farridav/django-jazzmin/issues/219


farridav:
@TysonRV do you have capacity to look through this ?

jazzmin/static/jazzmin/js/change_form.js
```diff
@@ -1,4 +1,7 @@
-function fix_selector_height() {
```
farridav, line 1: NO JS changes inside this PR except in jazzmin/static/jazzmin/js/ui-builder.js 

these diffs are just me wrapping everything in a closure, and passing jquery into it

tests/test_app/library/settings.py
```diff
@@ -1,5 +1,6 @@
 import os
 
+from django.conf.global_settings import LANGUAGES as DJANGO_LANGUAGES
```
TysonRV, line 3: Nice one this one!

Issue #225, opened 2020-12-05
Using `HiddenInput` in a custom form does not hide the label


matthewelwell:
When using a custom form on a model admin that overrides a specific field to use a hidden input, the label of that field is not hidden. Here is some basic code to demonstrate the issue: 

```python
# myapp/models.py

from django.db import models


class MyModel(models.Model):
    name = models.CharField(max_length=255)
    active = models.BooleanField()
```

```python
# myapp/forms.py

from django import forms


class MyModelForm(forms.ModelForm):
    active = forms.BooleanField(widget=forms.HiddenInput(), initial=True)
```

```python
# myapp/admin.py

from django.contrib import admin

from .forms import MyModelForm
from .models import MyModel


@admin.register(MyModel)
class MyModelAdmin(admin.ModelAdmin):
    form = MyModelForm
```

What I'd expect to see based on this code is the following: 

![image](https://user-images.githubusercontent.com/14089968/101266721-ae385600-3749-11eb-982b-7501750fc038.png)

but what I actually see is this: 

![image](https://user-images.githubusercontent.com/14089968/101266733-c4dead00-3749-11eb-8dc4-6a8bec061874.png)

With my limited CSS knowledge, the simple solution appears to be to add some styles to `jazzmin/css/main.css` for the `hidden` class, however, I'm not sure if there is a more elegant solution.

farridav:
I wonder why the input is hidden but not the label...  Perhaps we have done something wrong with the markup on field rendering

I'll try to reproduce this, and if I can fix the markup, that's the ideal resolution, failing that, your CSS fix is quite reasonable.

In the meantime you can add your CSS fix to a custom CSS file https://django-jazzmin.readthedocs.io/ui_customisation/#diy-with-custom-cssjs

matthewelwell:
Sure thing, thanks @farridav. Yep, I solved it with your suggestion already so no rush just thought I'd bring it to your attention.

farridav:
PR merged in :)

PR #226, opened 2020-12-05
Fix to ensure that hidden inputs are correctly hidden from admin screens

matthewelwell:
Fixes issue #225 in a possibly naive way by simply adding the `display: none` attribute to the `hidden` class.

There may be a more elegant solution to this problem so please forgive my naivety if that is the case.


farridav:
This seems like the best solution to the problem :) thanks

PR #227, opened 2020-12-07
Cleanup dark mode, update static files, bits and bobs

farridav:
- Cache tox dir for faster CI runs
- Try to fix github pages build failure (djangotemplate is not a valid Liquid tag)
- Add in note on how to customise CSS for browsers in dark mode
- Update adminLTE css and JS to latest
- make use of new dark-mode body CSS class in adminLTE
- fix indentation in JS code
- Add in more dark mode theme auto tweaks to UI builder
- run mypy checks during CI (needs some fixes)
- removes python 3.5 tests, as it has hit end of life already


farridav:
@TysonRV once this is in id like to get a release out, as theres already a few changes in master that need to go out ..

PR #228, opened 2020-12-07
Render fieldset name when using "single" template

bryan-brancotte:
Hi,

Thanks for the lib, it gives a nice look to the admin UI !

I encountered a small issue, fieldset names are not rendered when using single template. As a comparison, they are in vanilla Django, and also with other templates such as horizontal_tabs.

This PR add the name in a card-header when available and requested.

If anything is missing in this PR, please ping me back so I can do what is needed to make this PR mergeable.

Best regards


farridav:
Nice, I'll take a look at this first thing tomorrow, seems fine at first glance, but I'll take a look through all the change form variants .

PR #238, opened 2020-12-22
Update change_list.js

SalmanGill110:
'search-filter-mptt' class does not exist in template.  Causing filtering failure. 
Either change class in template or accept this.

### DIFF
--- a/jazzmin/static/jazzmin/js/change_list.js
+++ b/jazzmin/static/jazzmin/js/change_list.js
@@ -28,7 +28,7 @@
         //$ele.select2({ width: '100%' });
 
         // Use select2 for mptt dropdowns
-        const $mptt = $('.search-filter-mptt');
+        const $mptt = $('.search-filter');
         if ($mptt.length) {
             $mptt.search_filters();
             $mptt.select2({


farridav:
Change list mptt is there when the third party django app Mptt is used, I hadn't expected it to break when the app is not installed.


How does it break for you ?

SalmanGill110:
If you add filters to admin. Filters does not apply. 
choices variable always selects date:today as selected=true. I.e. changing values in filter drop down does not reach backend.

farridav:
I think it might be to do with this being commented out https://github.com/farridav/django-jazzmin/blob/master/jazzmin/static/jazzmin/js/change_list.js#L27

I'm not sure why that change was made, perhaps it was done by accident, I'll seenifni can get a hotfix in today, in the meantime you could put this into some custom js

```
$(document).ready(function () {
    const $ele = $('.search-filter');
    $ele.search_filters();
    $ele.select2({ width: '100%' });
});
```

I'll update this ticket once I've got a fix confirmed and shipped

SalmanGill110:
👍 makes sense..

farridav:
Closing in favour of https://github.com/farridav/django-jazzmin/pull/239

PR #239, opened 2020-12-22
Fix search filters

farridav:
- Ensure search filters are activated (Code was commented out?)
- Fix formatting on change list dropdowns (using `element` instead of `100%`)
- Fix some indentation


farridav:
Fixes: https://github.com/farridav/django-jazzmin/pull/238

Issue #195, opened 2020-10-31
Intermediate Grouping/Categorization of Menu Items


djpretzel:
Basically, I've got a "core" app with 40+ models, and I'd prefer not to break them up, but I'd also like to be able to create **arbitrary groups/categories for them in the sidebar menu**, and also have the option to insert those groupings (with dropdowns) along the top menu.

Right now I can hide specific apps (hide_apps), or models (hide_models), and I can add custom menu links to an app (custom_links), but I can't organize different models/admins *within* an app, which would offer a lot more flexibility.

**Additional/related ideas:** 

- ability to key custom links on model name, instead of app name.
- support for second/third-level menu items in the sidebar & dropdown
- given potential complexity/flexibility needs, move the entire menu to a dedicated class?
- a menu item could be either 'app', 'model', 'custom', or 'group', represented in a tree structure?

farridav:
I'll have a think about it, perhaps we can add an "advanced use cases" into the docs, and allow full control over the menu by overriding a class path or something..

Generally speaking, I'd highly recommend you split out your core app, though I respect there may be underlying reasons why that's not possible.

djpretzel:
@farridav Thanks, appreciated. At the moment this is a pretty big deal for me; I get that the Django paradigm pushes towards separation into different apps, but I also feel like it's optimistic to think that that separation can & should be tightly coupled to menu/navigation, and seeing Jazzmin support a workaround would be fantastic.

PR #241, opened 2020-12-22
Custom menu definition support

farridav:
Closes: https://github.com/farridav/django-jazzmin/issues/195

Aims to allow an entirely custom menu using an app-> model list mapping

e.g `"custom_menu": {"auth": ["books.book"], "any_name": ["auth.user", "auth.group"]}`

@djpretzel .. I have not forgotten about this, let me know if it supports your use case

TODO:
- [x] Finish test 
- [x] Update documentation


djpretzel:
@farridav Many thanks for remembering! Looks good, two questions:

1. Is it possible to nest an "arbitrary" (non-app) menu item under another one, or under a traditional app menu item for that matter? Right now the menu model is basically two levels, but this would allow for a third (when using the custom menu option) or even fourth, which I believe AdminLTE will support? In other words, envisioned as a tree, it would be nice to have these arbitrary groupings available not only at the very top level, but also one (or potentially 2/3) levels down - if that makes sense?

2. Will it cause issues for the "**model_str**" property if the "arbitrary name" text value is prefixed onto the model name, e.g. "**arbitrary name.user**" on line 213?

farridav:
1. No, currently nested groups are not supported. this would require additional work, if this PR adds value right now, id suggest we move forwards with it, and follow up with something that allows for nesting, as i see this ending up a lot more complicated

2. I have not spotted any issues so far, `model_str` is used for lookups in jazzmin config, e.g 

```
    "icons": {
        "auth.user": "fas fa-user",
        "Arbitrary Name.user": "fas fa-user",
        ...
      }
 ```

Supporting N levels of nesting is definitely something that needs scoping out

**Comment by vaz-lenbox** [comment]
Hi! I just discovered django-jazzmin and love it. This PR seems to be only feature that I am actually missing out on. Are there any plans to move forward with this ?

**Comment by farridav** [comment]
I've got a building list of things to take care of here, and my other maintainer is currently unavailable, so it's been tricky getting it all done, along with my other work.

This ticket is definitely one I want to get finished up, hopefully within the next few weeks if possible.

**Comment by 1171000408** [comment]
I wonder if it is possible to add a apps group design in this feature as it is quite necessary for a project with a series of apps.
I'm working with a project with a design of 12 apps devided into two different groups, and I really feel messy to look through the side menu.
Using prefix in appconfig verbose name and order_respect_to is a replacement solution but my colleague said it was a really stupid way to deal with this problem hah :D.

**Comment by Pijuli** [comment]
It's not a push, but just a question.
How is this feature going? Can anyone give some help to get it finished?
Thank you so much!

**Comment by chaleyct** [comment]
Hey just checking to see if this feature still has a path forward. Thanks for the hard work!

**Comment by farridav** [comment]
Yes this is still definitely something I want to get to, I'll try to set
some time asides over the next couple of weeks to get it finished off, it's
been sitting here for ages ... :(

File: jazzmin/templatetags/jazzmin.py [hunk 1]
```diff
@@ -35,13 +35,34 @@
...
     order_with_respect_to,
+    regroup_apps,
 )
 
 User = get_user_model()
 register = Library()
 logger = logging.getLogger(__name__)
 
 
+def _process_models(app_label: str, models: List[Dict], options: Dict) -> List[Dict]:
```
katiam2, line 46: This is a simple extraction to ruff requirements.

katiam2:
Already implemented

Muhammadinaam:
@katiam2 , thanks if it has already been implemented. Please tell me how to use this feature or refer me to the documentation link. Thanks for working on such a nice package.

Issue #193, opened 2020-10-30
Support for django-admin-numeric-filter (feature)


hwalinga:
Django-admin-numeric-filter (https://github.com/lukasvinclav/django-admin-numeric-filter) adds support for adding range filters on your lists with some nice sliders. 

However when using this with django-jazzmin you are greeted with:

```
....azzmin/templatetags/jazzmin.py", line 234, in jazzmin_list_filter            
query_string = choice["query_string"][1:]                                                
KeyError: 'query_string'                                                                            
```

This works without problems in the default admin.

I think jazzmin makes assumptions about how filters work which is broken by django-admin-numeric-filter. 

Would be nice if the page does not break and maybe the range slider box needs some jazz.

farridav:
When i look into the search and filter fields implementation, i feel like it needs reshuiffling, i think its drifted away from djangos implementation too much, to fix this issue, i think we would rework the `search_form.html` include in `change_list.html`, and if possible, do away with the `jazzmin_list_filter`

The fastest way for us to provide a fix in for this, is to do the following:

1. Fork the project and make a new branch
2. Integrate this 3rd party app into our embedded demo app (https://github.com/farridav/django-jazzmin/tree/master/tests/test_app)
3. push it up and send us a link to the branch
  
This way we will be able to reproduce the issue quickly, and fix it, without needing to delve into how to integrate/setup the 3rd party app :) 

Thanks for reporting, we hope to provide compatibility with all other packages, especially if they are widely used

hwalinga:
I created a branch where I integrated the 3rd party app. See PR. 

Good news: It just works now, but only needs some styling.

arneatec:
Bump as django-admin-numeric-filter is still missing styling.

farridav:
I'm afraid I'm finding it very difficult to put time in to this at the moment, so anybody that wants to add to the PR, and get it code complete, please do. 

I hope to have more time in a month or so, but until then, I'll need contributors

PR #251, opened 2021-01-18
WIP: Support django-admin-numeric-filter

hwalinga:
This adds django-admin-numeric-filter to the test app. It does this by adding a pages field to the Books and add it to the admin.

It works, but it just needs some styling. 

See #193


hwalinga:
Seems the CI does not install the deps specified in the pyproject.toml.

farridav:
Yeah they also need to be in project.toml as tox takes its config from there ...

hwalinga:
>Yeah they also need to be in project.toml as tox takes its config from there ...

Eh, not experienced with tox, but can't you just put `poetry install` in the commands tox executes?

This might give you more pointers: https://github.com/python-poetry/poetry/issues/1745

hwalinga:
Unsure what this means:

```
  E         {'django/forms/widgets/input.html': 26} != {'django/forms/widgets/input.html': 21}
  E         {'django/forms/widgets/attrs.html': 32} != {'django/forms/widgets/attrs.html': 27}
  ```

**Comment by farridav** [comment]
It means that those widgets have been rendered out 5 more times than expected.

These assertions help us to ensure we don't accidentally put includes in loops, and cause the page to take ages to render.

**Comment by hwalinga** [comment]
Ah, so because I changed the admin, there is now more to render, and that failed the tests? Should I increment those than so that the numbers match?

This does look a bit prone to breaking. Every time someone adds something to the admin, those numbers aren't correct anymore. 

Also, do you want in this PR an example usage of #244 as well. That is essentially the exact same problem, but just another third party app.

**Comment by farridav** [comment]
Sure, you can bump the numbers.

It does require maintenance, though it also prevents error, so it's adding value right now

**Comment by hwalinga** [comment]
Okay, but do I also need to add an example for #244 then?

**Comment by farridav** [comment]
Sure that's fine, it's mainly to help us verify the 3rd party app support works as expected.

If the demo app gets too full of 3rd party apps, and obscures the jazzmin implementation we will clean it up.

**Comment by hwalinga** [comment]
Your CI discovered a funny bug:

```
  >   date_of_death = factory.LazyAttribute(lambda x: x.date_of_birth.replace(year=NOW.year))
  E   ValueError: day is out of range for month
```

(leap year)

Fix in: #258

**Comment by hwalinga** [comment]
So this includes now correctly the two third party apps. Now someone needs to look into the CSS to make this look good. I think you have a much better understanding of CSS and bootstrap, but if you don't have time I don't have any problems looking into it myself.

**Comment by hwalinga** [comment]
@farridav Friendly ping, as this is gathering merge conflicts.

Should I see if I can fix the CSS to support this, or do you want to take a look? Or should we first merge this already, and in a later PR fix the CSS to add support for it?

File: pyproject.toml [hunk 1]
```diff
@@ -43,6 +43,8 @@ exclude = ["tests", "docs"]
 [tool.poetry.dependencies]
 python = ">=3.5"
 django = ">=2"
+django-admin-numeric-filter = "^0.1.6"
```
farridav, line 46: We dont want to add these here, as it will force everyone who uses jazzmin to have them installed ..

best to put them under the dev dependencies see(`# 3rd party app support demonstrated in test app`)

tests/test_app/library/settings.py
```diff
@@ -18,6 +18,8 @@
 INSTALLED_APPS = [
     # Keep this above 'django.contrib.admin'
     "jazzmin",
+    "admin_numeric_filter",
```
farridav, line 21: Best we add a comment here, something like

`# 3rd/4th party app support demonstrated by jazzmin, NOT needed for use of jazzmin`

farridav (changes requested):

farridav:
If you can take a look at the CSS that's a win, I'm a bit overloaded with other projects at the moment, so tend to be looking at the easy wins on the weekends.

I've left some comments on the PR that we will need to address first.

If the styling is currently OK, we can get it into master, and follow up with jazz as needed

farridav:
Getting this into mainline, as it provides support, and no breaking changes for those not using the 3rd party app.

But let's follow up with some style improvements asap ..

Issue #249, opened 2021-01-17
AdminSite subclasses and jazzmin_settings


olof:
Hello, thanks for a really nice project! Very nice look for the admin interface. Real jazzy!

I just looked at enabling it for my django project, using subclassed AdminSites to expose different views to different user classes. Just adding it to INSTALLED_APPS, everything works fine and looks really good.

But one thing I noted when using it with my use case: the title reverts to django admin's default value. In the `base.html` template, `{{ jazzmin_settings.site_title }}` and `{{ jazzmin_settings.site_header }}` are used, rather than what django.admin exports in the template context (i.e `{{ site_title }}` and `{{ site_header }}` respectively). In my AdminSites these properties are overriden, but this isn't reflected in jazzmin_settings. Changing the template to just using `{{ site_title }}` in this case works. I recognize that simply doing that wouldn't be backwards compatible, but is there a way to handle my use case somehow?

As I said, without any configuration, it otherwise works really well. But were I to configure jazzmin, the documented way is through my django settings.py; is there a way to override settings per AdminSite class?

Thanks a lot!

farridav:
perhaps there is a way we can get the  current admin site from the request, and pull the site title and site header from that (unless it is specified in `JAZZMIN_SETTINGS`

We have limited support for use outside of the default admin site, but would like more .. 

if your able to reproduce the problem in our [demo app](https://github.com/farridav/django-jazzmin/tree/master/tests/test_app), and link me to it then ill have a go at sorting it ..

olof:
Thanks! I'll take a look!

olof:
I've made changes to the test app to demonstrate the issue, the commit is available at:

https://github.com/olof/django-jazzmin/commit/bd77566e0adcd7479227351dc3ac3175e6647c21
https://github.com/olof/django-jazzmin/tree/gh/249/custom_admin_site/tests/test_app

While retaining the old admin site, it also adds a "limited" view for librarians at `/en/staff/`. Here, the issue of jazzmin not picking up the custom AdminSite properties can be seen (most notably in the top left corner). (I've commented out the `site_title` and `site_header` settings in `settings.py` to make the issue stand out a bit more, it now simply says "Django administration"). Hope this is along the lines of what you asked for, otherwise let me know.

PR #253, opened 2021-01-21
Default site title and header from current admin site

farridav:
Use the "current" admin site for site title and header defaults, this has to be determined from the request

Fixes https://github.com/farridav/django-jazzmin/issues/249

@olof 

Let me know if this resolves your issue, i was able to confirm it working using your commit


olof:
Hello, I've tried it out with my real django project and it does indeed work well! Thanks for taking the time and addressing my problem so quickly.

farridav:
No worries, thanks for helping me reproduce the problem

PR #256, opened 2021-02-26
Set dark_mode_theme to src for the dark mode.

### LINKED ISSUE #255: Bug that prevents dark-themes from being applied.
I think I found a bug.

The line 283 in the file jazzmin\settings.py:
`        "dark_mode_theme": {"name": dark_mode_theme, "src": static(THEMES[theme])},
`
need to be replaced with:
`        "dark_mode_theme": {"name": dark_mode_theme, "src": static(THEMES[dark_mode_theme])},
`

The problem is that in the current version there is no case where a dark-theme stylesheet can be loaded.

hwalinga:
Closes #255


hwalinga:
Closing in favor of @257

PR #257, opened 2021-02-26
Set darkmode theme on src

farridav:
Closes: https://github.com/farridav/django-jazzmin/pull/256

Thanks @hwalinga i see what you mean, ive just changed it so that it doesnt raise the `KeyError`


hwalinga:
Closes issue #255

PR #261, opened 2021-03-11
fix error style fo field in tabular_inline

dimastbk:


farridav:
Looks good, I'll check this out and get it in asap

PR #278, opened 2021-04-02
Compatible for py3.8

dongyi:
compatible for py3.8


dongyi:
I made this change for running in py3.8, it raised an error before **TypeError: can't pickle generator objects for Non Generator methods**

farridav:
perhaps this is caused by django using a generator for `context['available_apps']` as opposed to a list, we already test on python3.8 and am fairly confident we have coverage here..

We can get this in, as it seems to resolve some combination of django & python, but im not so sure which...

Here is our current list of test environments https://github.com/farridav/django-jazzmin/blob/master/pyproject.toml#L104

Issue #283, opened 2021-04-07
Possibly mismatched permission


sondrelg:
Hi! 

First of all thanks for maintaining this package - it's really good 👏 

We recently decided to try and implement `django-jazzmin` in one of our projects, and after the switch we got a report from a user that they could no longer see the `History` button on a model detail views. After looking through the code I've managed to boil it down to [this line of code](https://github.com/farridav/django-jazzmin/blob/master/jazzmin/templates/admin/solo/change_form.html#L14). Seems you need one of two permissions or the button is hidden.

Fixing this was easy enough, but after toggling off the theme, I see that this doesn't in fact match the Django admin. I tried finding the Django equivalent, and think [this](https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/change_form_object_tools.html#L4) is the original link. 

I suppose this *might* be intentional, but in the case that it isn't I thought I'd let you know 🙂

farridav:
The plan is to match Django in all areas possible, so if the button is showing in Django, but not with jazzmin, then there is a fix needed on our side.

If you wanted to submit a PR for this, I'll take a look, otherwise I hope to do some bits towards the end of the week.

Thanks for reporting :)

farridav:
No worries, I'll update this ticket once I have it resolved

sondrelg:
Actually I guess the fix is easy enough to set up for you, so I opened a quick PR 🙂

PR #284, opened 2021-04-08
Remove permissions check for History button to match Django

sondrelg:
Resolves https://github.com/farridav/django-jazzmin/issues/283

I can't see any permissions required for the `History` section in Django admin, and with `django-jazzmin`, navigating to `/history` works. This PR removes the conditional button rendering to (hopefully) match Django 🙂


farridav:
Makes sense to me, not sure why we checked for change permission in the first place, as history doesnt allow you to change anything.

PR #285, opened 2021-04-08
add event listener resize for GeoDjango Pointfield map

### LINKED ISSUE #265: Problem with map on tabbed model view with Point
Hello,
first of all, thank you for all work that was put into this project!

We found a problem with `PointField` in Admin.

### Prerequisites

- Django: 3.1.7
- Jazzmin: 2.4.4
- Python: 3.8
- Chrome: 88.0.4324.190

## What is the problem:
When you use tabbed view together with `PointField `, not on the first page, the map will be blank till you make any screen resize or zoom:
image
When you resize the window/zoom in/out, the map appears!
image

**What works:**
- Not tabbed view
- Map on the first tab

**Additional info:**
HTML when map is not visible:
`<canvas class="ol-unselectable" style="width: 100%; height: 100%; display: none;"></canvas>`   
HTML after page resized:
`<canvas class="ol-unselectable" style="width: 100%; height: 100%;" width="600" height="400"></canvas>`

## How to reproduce:
**You can use this repo made just for this: https://github.com/Fasther/jazzmin-point-bug/tree/master/jazzmintest/testj**
Needs Django, Jazzmin and PostGIS installed...

_or..._
**Step by step:**
1. Create a blank Django project and app
2. Define model with `PointField` and some other field (doesn't matter)
3. Create admin, where you will place the map on the second group/tab
4. In admin after you switch to this tab the map will be blank

kainbacher:
We have a project where we use the PointField from GeoDjango. If the map is on the second tab it's size is 0px.
To fix this we call a document resize and the map gets also resized.

Maybe the same issue can happen with some other fields.

fix #265

### DIFF
--- a/jazzmin/static/jazzmin/js/main.js
+++ b/jazzmin/static/jazzmin/js/main.js
@@ -41,6 +41,15 @@
         }
     }
 
+    function callResizeOnTabSwitch() {
+        /*
+        used to reinitialize the map on a GeoDjango Pointfield
+        */
+        $('a[data-toggle="pill"]').on('shown.bs.tab', function (event) {
+        window.dispatchEvent(new Event('resize'));
+        });
+    }
+
     $(document).ready(function () {
         // Set active status on links
         setActiveLinks()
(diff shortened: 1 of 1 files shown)


farridav:
Ill get this in for now, but will follow up with a change that moves this into change_form.js, and applies it to the other change form templates too

PR #293, opened 2021-04-14
Bumped python version to 3.6, added black

dhvcc:
- Added `black` to dev dependencies
- Changed min python version to `3.6`
- Removed `.python_version` file that was fixing version for local development

### Why?

1. This project does not test `python3.5`
2. `python3.6` is minimal for `black` to work (may be the reason why `3.5` is not being tested)
3. `.python_version` seems to be an unnecessary file which locks python to `3.9.4`. This should be also ignored I guess
4. `black` is used in `Makefile`, but not listed in project's dev deps (when `flake9` and `mypy` are)


farridav:
Looking good, I'll follow up with adding .python-version to .gitignore

dhvcc:
> Looking good, I'll follow up with adding .python-version to .gitignore

Ok, nice 🙂

PR #306, opened 2021-04-30
Add correct classes to django actions html

matthewelwell:
I noticed that the 'select all x models' functionality wasn't working (I'm running Django 3.2). From digging into it, it looks like it's related to the use of the `hidden` class over the inline styles which was changed in Django 3.2 as per the commit here: https://github.com/django/django/commit/30e59705fc3e3e9e8370b965af794ad6173bf92b.

As I'm writing this, I think the change that I've suggested here, however, would mean this functionality would not work correctly for versions of django prior to 3.2 - I'll have a think about whether there's a better solution but please let me know your thoughts if you have time.


farridav:
Makes sense to me

PR #314, opened 2021-06-10
Fixed conflict with other iframes in add/change form

### LINKED ISSUE #313: Conflict with other iframes in add/change form
Add related object modal is not working as expected when there's other iframe in the add/change form.

It's not the best idea to depend on the number of `iframe` tags. I suggest using class name (`related-iframe`) instead.

lchojnacki:
Resolves #313


farridav:
Makes a lot of sense, thanks

PR #315, opened 2021-06-11
Fixed checkIfInIframe and initialized related modal for formsets

lchojnacki:
1. This issue is similar to #313, we should not get elements by tag name. It's better to use class name instead :)
2. Modal was not being initialized in inline formsets with `raw_id_fields` (screenshot). In this case the content of modal was opened in new window.
inline_formset_raw_id_fields


farridav:
Makes sense, good to see

PR #316, opened 2021-06-12
Changed wrong variable name at cli.py

minholi:
Tried to use this script but it was not working, so I did this and now it works fine.


farridav:
Aah yes, your right, the argument should have been named prune.

Thanks

PR #331, opened 2021-08-27
Feature/filter input length

### LINKED ISSUE #319: Filter on M2M is incredibly slow
When using a filter on M2M with like 20k row the filter is **so slow**
Would be nice not to re-render dropdown items on each click which is happening right now

I was trying to use 
```css
.select2-results__options li:nth-child(n+10){
  display: none;
}
```
To not render all 20k fields, but it's no help because every item is inserted with javascript at runtime

Dropdown is cool and it's search if fast, but the initial "render" of dropdown takes just too long. We should probably handle this in template

dhvcc:
A feature to add `filter_input_length` class attritube for admin models to change select2's `minimumInputLength` (which defaults to 0)

- [X] Implementation
- [x] Docs
- [x] Questions resolved

closes #319


dhvcc:
@farridav :eyes:

farridav:
As soon as I'm back at my computer I'll take a look through this, hopefully with 24hrs :)

jazzmin/templates/admin/change_list.html
```diff
@@ -111,5 +111,14 @@ <h4 class="card-title">{{ title }}{% block pretitle %}{% endblock %}</h4>
 {% block extrajs %}
     {{  block.super }}
     <script type="text/javascript" src="{% static 'vendor/select2/js/select2.min.js' %}"></script>
+    <script>
+        {% comment %} set filterInputLength default and custom values {% endcomment %}
+        window.filterInputLengthDefault = 0;
+        window.filterInputLength = {
```
farridav, line 117: Is there value in being able to set the minimum input length per model admin ? Or would it simplify things to have it set within `settings.JAZZMIN_SETTINGS` ?,

dhvcc: > It's per admin model. Some filters may have 20K+ records and some just 5. I think we can use global setting to provide a default, and then customize it if needed

farridav:
This all looks good, though we will need to update the documentation to tell people they can use the `filter_input_length` property on model admin classes

dhvcc:
@farridav will do later this week. So should we implement a setting to change the default 0 input length

dhvcc:
I decided to not implement global default for this feature, because it seems unnecessary and the current default of 0 seems pretty logical

farridav:
LGTM

PR #332, opened 2021-08-30
split brand from header

SourceDoctor:
remove "elevation-3" to avoid creapy shadowing on logo

split site_header to site_header  and site_brand

site_header is on login_screen
site_brand  is Title on brand

so brand can be reduced to logo only, with empty string, or short name of brand


SourceDoctor:
@farridav ?

farridav:
This makes sense to me also

PR #341, opened 2021-09-20
Highlight, and expand active app when on detail view

saeedhmt:
_Edited:_

When going to the detail view of a model, find the parent app group in the sidemenu, highlight and expand it e.g@

image

Nothing selected:

image

### DIFF
--- a/jazzmin/static/jazzmin/js/main.js
+++ b/jazzmin/static/jazzmin/js/main.js
@@ -56,7 +56,14 @@
         const $changeListTable = $('#changelist .results table');
         if ($changeListTable.length && !$changeListTable.hasClass('table table-striped')) {
             $changeListTable.addClass('table table-striped');
-        }
+        };
+
+        var a_active = $('a.nav-link.active');
+        var main_li_parent = a_active.closest('li.nav-item.has-treeview');
+        var ul_child = main_li_parent.children('ul');
+
+        ul_child.show();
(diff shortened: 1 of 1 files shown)


jazzmin/static/jazzmin/js/main.js
```diff
@@ -56,7 +56,14 @@
         const $changeListTable = $('#changelist .results table');
         if ($changeListTable.length && !$changeListTable.hasClass('table table-striped')) {
             $changeListTable.addClass('table table-striped');
-        }
+        };
+
+        var a_active = $('a.nav-link.active');
+        var main_li_parent = a_active.closest('li.nav-item.has-treeview');
+        var ul_child = main_li_parent.children('ul');
+
```
farridav, line 64: please change `var`'s to `const` and use `$` prefixes for constants that contain jquery objects

farridav:
This looks good :) one small change to make, then happy to get this in.. thanks for your contribution

jazzmin/static/jazzmin/js/main.js
farridav: Maybe also move this code up into the `setActiveLinks` function

saeedhmt: done

saeedhmt:
> This looks good :) one small change to make, then happy to get this in.. thanks for your contribution

PR #342, opened 2021-10-01
show fieldset description

SourceDoctor:
fieldset definition can have "description" field beside "fields" to describe those "section"

https://docs.djangoproject.com/en/3.2/ref/contrib/admin/

"
A string of optional extra text to be displayed at the top of each fieldset, under the heading of the fieldset. This string is not rendered for TabularInline due to its layout.
"


farridav:
Without screenshots, I'll need to check this one out and see how it renders.

My only question would be whether it needs to be in a p tag as opposed to a div.

I'll try to check this out later on today.

PR #348, opened 2021-11-04
Fix(Padding): fix layout padding in table

Laityned:
Fix padding in table, which is a bug in adminLTE:

jazzmin table incorrect padding

To:

jazzmin table correct padding

See column of username


farridav:
Nice ! Thanks for your contribution

PR #351, opened 2021-11-17
Make Google CDN usage optional for fonts

### LINKED ISSUE #347: Dependency on Google Fonts
Currently there is a hard coded dependency on fonts.googleapis.com in the base template.
https://github.com/farridav/django-jazzmin/blob/a69840caa7639daee8400710a877038efa1a7cc1/jazzmin/templates/admin/base.html#L44

This seems to be the only such dependency, as all other third-party resources (FontAwesome, etc.) are served locally. In a project I'm developing we are not allowed to access that external domain. While it's possible to override the base template, this seems quite overkill to change a single URL, and requires us to keep up with upstream changes to that file.

Would it be possible to make this font URL configurable, such that one can select a locally served font resource instead?

RenWal:
This adds a setting `use_google_fonts_cdn` that allows the server admin to switch off the automatic include of font resources from fonts.googleapis.com . In this case, `custom_css` can be used to supply fonts. This allows changing the font in use, or simply serving the Google font locally.

Closes #347

### DIFF
--- a/jazzmin/settings.py
+++ b/jazzmin/settings.py
@@ -74,6 +74,8 @@ DEFAULT_SETTINGS: Dict[str, Any] = {
     # Relative paths to custom CSS/JS scripts (must be present in static files)
     "custom_css": None,
     "custom_js": None,
+    # Whether to link font from fonts.googleapis.com (use custom_css to supply font otherwise)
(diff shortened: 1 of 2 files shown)


farridav:
This seems reasonable to me

PR #353, opened 2021-12-02
Using method in user avatar

joearton:
When function is on User class, the below code can't be reached.

```
def get_profile_picture(self):
    if self.property.picture:
        return self.property.picture.url
    if hasattr(self.app_user, 'picture'):
        if self.app_user.picture:
            return self.app_user.picture.url
    return None
User.add_to_class('get_profile_picture', get_profile_picture)
```


farridav:
This seems quite reasonable

PR #354, opened 2021-12-02
Remove pr-2 from user and language selector to fix uneven right size of the focus pseudo-class

Pijuli:
This does not solve any open issue but I think is a fix to take in account.

This fix is meant to fix the uneven padding on the top right user and language buttons.

It might also be addressed by changing to px-1 instead of removing pr-2, in case I'm missing any reason for it to be there

Hope this small grain of sand helps


farridav:
This makes sense to me :)

PR #377, opened 2022-03-10
Fix missing button in StackedInline

### LINKED ISSUE #376: [Styling bug] Stacked inline - missing inline "button"
For `TabularInline; it looks like this:
image

For `StackedInline` like this.
image

Should be:
image

Fasther:
Fix #376

Looks like this after repair.
image


Fasther:
I am not JS pro, so feel free to edit.

farridav:
This seems reasonable

PR #381, opened 2022-03-17
Remove and ignore pyenv's local python version

dhvcc:
Since this project supports multiple versions we ideally don't want to pin python's version with pyenv


dhvcc:
@farridav may be you could take a look and tel me what's wrong with builds? They seem to fail for no reason (updated setup-python version after [this fail](https://github.com/farridav/django-jazzmin/runs/5593592905?check_suite_focus=true))

farridav:
Sure, I'll take a look at this today

farridav:
Im not entirely sure why this failed, it seems that flake8 (the flake9 fork) was not picking up the pyproject.toml config, so didnt adjust its line limit from the standard 79 chars.

Ive added a new commit on top of this PR: https://github.com/farridav/django-jazzmin/pull/382

and its passing, ill close this PR and merge that one in

PR #383, opened 2022-03-23
Fix for related objects inline edit/delete buttons

dekomote:
After select2 is applied to the related objects select inputs, the event
that enables/disables the edit and delete buttons next to the select is
not being triggered. Which makes it impossible to click those buttons
and open the popups
This code rebinds the change event after applying select2.


farridav:
Seems sane to me, thanks :)

PR #345, opened 2021-10-17
Added Reset Password in admin

ghost:

PR #389, opened 2022-03-26
Support password reset views

### LINKED ISSUE #387: Login template is missing link to password reset
The [documentation of django admin](https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#adding-a-password-reset-feature) claims:

> The presence of the admin_password_reset named URL will cause a “forgotten your password?” link to appear on the default admin log-in page under the password box.

So I guess django-jazzmin should follow that principle and include a similar block like in the [default login template](https://github.com/django/django/blob/abfdb4d7f384fb06ed9b7ca37b548542df7b5dda/django/contrib/admin/templates/admin/login.html#L56-L61).

_Note: There is already an open PR for this (https://github.com/farridav/django-jazzmin/pull/345), but the author seems to have deleted their GitHub account._

### LINKED ISSUE #388: Password reset views use unsuitable templates
When the default password reset views are enabled, the same templates are used as for the authenticated admin views:

jazzmin-password-reset-form

jazzmin-password-reset-confirm

Alternatively, I suggest to use the same basic box-layout as for the login form.

timoludwig:
Thanks for your project! We really like the design, however we miss the possibility to reset passwords for our users.

Since PR #345 did not cover the complete password reset process and seems to be abandoned, I decided to take a shot at this topic.

### Suggested changes:
- Generalize login template (make it possible to reuse the design of the login form for other non-authenticated views)
- Add link to password reset to login form (fixes #387)
- Add templates for password reset (fixes #388)
- Add tests for password reset views/templates

@farridav Thanks in advance for your feedback! In PR #345, you suggested to add a setting for the name of the password reset view, however in the [default template](https://github.com/django/django/blob/abfdb4d7f384fb06ed9b7ca37b548542df7b5dda/django/contrib/admin/templates/admin/login.html#L56-L61), the view name is hardcoded as well, so I think it's fine to adopt this for django-jazzmin. If you still think a setting is required, I can add this to this PR if you want.

### DIFF
--- a/jazzmin/templates/admin/login.html
+++ b/jazzmin/templates/admin/login.html
@@ -55,6 +55,16 @@
                 </div>
             </div>
         </div>
+        {% url 'admin_password_reset' as password_reset_url %}
+        {% if password_reset_url %}
+            <div class="mb-3">
+                <div class="password-reset-link" style="text-align: center;">
+                    <a href="{{ password_reset_url }}">
+                        {% trans 'Forgotten your password or username?' %}
+                    </a>
+                </div>
+            </div>
+        {% endif %}
         <div class="row">
             <div class="col-12">
                 <button type="submit" class="btn {{ jazzmin_ui.button_classes.primary }} btn-block">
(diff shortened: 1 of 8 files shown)


farridav:
Nice job! it all seems to play nicely, and I appreciate you adding a test in to cover the functionality, ill get this merged and released ASAP

PR #392, opened 2022-04-02
fix: fixes ModuleNotFoundError: No module named 'dj_database_url'

PrynsTag:
Running `python tests/test_app/manage.py migrate` shows **ModuleNotFoundError**.

On system:
Ubuntu 20.04
python 3.10
Poetry version 1.1.13

### DIFF
--- a/jazzmin/templates/jazzmin/includes/carousel.html
+++ b/jazzmin/templates/jazzmin/includes/carousel.html
@@ -25,7 +25,7 @@
     </ol>
     <div class="carousel-inner">
         {% for fieldset in forms %}
-            <div class="carousel-item {% if forloop.first %}active{% endif %}" data-carouselid="{{ forloop.counter0 }}" data-label="{{ fieldset.name|default:general_tab|capfirst }}" data-target="#{{ fieldset.name|default:general_tab|unicode_slugify }}-tab">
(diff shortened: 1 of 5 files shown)


farridav:
I'm afraid this will force everyone installing jazzmin to have this dependency, all of the dependencies used by the test app are defined in [tests/test_app/requirements.txt](https://github.com/farridav/django-jazzmin/blob/master/tests/test_app/requirements.txt)

I usually `pip install -r tests/test_app/requirements.txt` from within my poetry environment, though i understand thats a little counter intuitive, and currently undocumented.

We could either document this in https://django-jazzmin.readthedocs.io/development/ and/or duplicate the dependencies to the dev packages section in [https://github.com/farridav/django-jazzmin/blob/master/pyproject.toml#L47][pyproject.toml]

Those dependencies are there because the heroku test app reads that file on deployment to install the packages.

PrynsTag:
It now works when I ran the command: `pip install -r tests/test_app/requirements.txt`. And you are right. This is not documented and for new contributors like me, it is easy to miss. 

I suggests we do the minimal implementation of the solution which is to document the extra steps to install required package. However, placing it in dev-dependency makes much more sense.

I am closing this PR.

PR #398, opened 2022-04-22
Use fixed width for registration templates (fixes #318)

### LINKED ISSUE #318: login form is expanded when invalid login credentials are provided
When entering wrong credentials for login, the form is disturbed and expanded and it looks ugly. How can I fix this ?

timoludwig:
I'd also prefer a fixed width for the registration/login box, independently of the help text or error messages.

Fixes: #318


farridav:
I agree, it does look better keeping it a consistent width, and seems to render out just fine on mobile too

PR #399, opened 2022-04-22
Use registration base template for logout

timoludwig:
I think it makes sense to use the same base template as for all other "unauthenticated" views...


farridav:
Nice, this drys things up a bit :)

PR #401, opened 2022-04-23
Workflow updates

dhvcc:
1. Set `cache: pip` everywhere to speed up workflows a bit
2. Base actions version bump
3. Fix `3.10` version in lint workflow to use `latest` (and cached) version of python

Adding pip cache will result in less traffic on PyPi (but at this scale pertty much won't speed up perfomance). Same for using latest and cached by GitHub version of 3.10

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,5 @@
 ---
+
 name: Build
 on:
   push:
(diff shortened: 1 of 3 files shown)


farridav:
Looks good, thanks for the optimisation

PR #405, opened 2022-04-27
Styling/basic-cleanup

kainbacher:
in this pr i have done a lot of smaller changes. basically styling things. the idea is to make jazzmin a bit cleaner.


farridav:
Looks nice so far, though i encourage you to switch your theme to darkly and take a look at it there, it might need some tweaks (a lot of people opt for dark themes either explicitly, or via browser/OS preference)

Heres an image of how it looks for me:
jazzmin_dark_mode

farridav:
Thanks, this adds some nice bits, shame you had to remove the grey link colours, as that looks pretty smart on the default light themes.

I plan to follow up with some changes on improving dark themes, as there are some bits that need fixing outside of this PR.

I have found that the best way to keep things responsive, consistent and compatible with all of the bootswatch themes (including the dark ones) is to avoid CSS overrides wherever possible, and try to fix the markup so the existing CSS selectors work as intended, as we are not currently making use of a css pre-processor and hope to get away with:

- AdminLTE base (which is compiled with bootstrap)
- Optional theme (using bootswatch)
- Essential css overrides to fix django or adminLTE specific things (this needs whittling down, its currently bloated)
- Optional user DIY CSS

Issue #404, opened 2022-04-27
Branding is not applied in admindocs if site_brand is not defined in jazzmin settings


byenilmez:
Hi, thanks for the great theme, it make django look more modern.

I have come across a bug while using this theme with admindocs.
The template tag does not pick up admin app values when request is coming from admindocs. This results in showing "None" in admin docs when branding is not defined in jazzmin but on admin.site.

I have modified my local copy as such which seems to fix the problem:

> jazzmin/templatetags/jazzmin.py

```
@register.simple_tag
def get_jazzmin_settings(request: WSGIRequest) -> Dict:
    """
    Get Jazzmin settings, update any defaults from the request, and return
    """
    settings = get_settings()

    admin_site = {x.name: x for x in all_sites}.get("admin", {})
    if not settings["site_title"]:
        settings["site_title"] = getattr(admin_site, "site_title", None)

    if not settings["site_header"]:
        settings["site_header"] = getattr(admin_site, "site_header", None)

    if not settings["site_brand"]:
        settings["site_brand"] = getattr(admin_site, "site_header", None)

    return settings

```

PR #406, opened 2022-04-28
Fix admindocs branding if it's using values from admin

byenilmez:
A fix for Issue #404

### DIFF
--- a/jazzmin/templatetags/jazzmin.py
+++ b/jazzmin/templatetags/jazzmin.py
@@ -128,16 +128,15 @@ def get_jazzmin_settings(request: WSGIRequest) -> Dict:
     """
     settings = get_settings()
 
-    if hasattr(request, "current_app"):
-        admin_site = {x.name: x for x in all_sites}.get(request.current_app, "admin")
-        if not settings["site_title"]:
-            settings["site_title"] = admin_site.site_title
+    admin_site = {x.name: x for x in all_sites}.get("admin", {})
+    if not settings["site_title"]:
+        settings["site_title"] = getattr(admin_site, "site_title", None)
 
-        if not settings["site_header"]:
-            settings["site_header"] = admin_site.site_header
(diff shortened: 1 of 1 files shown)


jazzmin/templatetags/jazzmin.py
```diff
@@ -128,16 +128,15 @@ def get_jazzmin_settings(request: WSGIRequest) -> Dict:
     """
     settings = get_settings()
 
-    if hasattr(request, "current_app"):
-        admin_site = {x.name: x for x in all_sites}.get(request.current_app, "admin")
-        if not settings["site_title"]:
-            settings["site_title"] = admin_site.site_title
+    admin_site = {x.name: x for x in all_sites}.get("admin", {})
```
farridav, line 131: if the default value for this `.get` is a dictionary, then we should probably not be using `getattr` on it.

Also, how comes you have removed the `request.current_app` key ?

It would be good to know what error you are receiving, as i think we might be able to find a more compatible solution

byenilmez: You are right. Setting default value to `dict `and using `getattr `is confusing. We've used `getattr `since, if defined, `admin_site `should be instance of `AdminSite `which does not have a get method. It might be better to remove the default value to make it `None` when not available, since `getattr `can handle `None`.

We've removed the `request.current_app` since it is only defined when `admin `app is the current app. So it doesn't exists when browsing admindocs pages. We can keep it and set default to "admin" as before. We wanted to be sure if app name in admindoc is defined later on it won't pick up adminsdocs name.

Btw, we did not receive any errors, but the site_title showed up as None:
image

We have defined title via `site_header `in out custom admin class instead of jazzmin settings. If it's defined in jazzmin settings, this error is not replicated as it gets everything from `get_settings()` properly.

farridav: Apologies for not getting this in sooner

PR #409, opened 2022-05-20
Set default_app_config only in the appropiate Django versions

agomez-currencyalliance:
Otherwise adding jazzmin to a Django 4.0 and running it with deprecation warnings throws the following warnings
`RemovedInDjango41Warning: 'jazzmin' defines default_app_config = 'jazzmin.apps.JazzminConfig'. Django now detects this configuration automatically. You can remove default_app_config.`


agomez-currencyalliance:
@farridav I'm sorry, but would you mind deleting this pull request using that link? I contacted you on email too.

Issue #362, opened 2022-01-05
List page first table column too narrow if a user is read only


Blaza:
If a user is read only, the checkbox for selecting all the models is not shown by django because the user can't do any actions.
In the code here:
https://github.com/farridav/django-jazzmin/blob/a69840caa7639daee8400710a877038efa1a7cc1/jazzmin/templatetags/jazzmin.py#L437
the class `djn-checkbox-select-all` is added and that makes the first column narrow. Removing that class from the `th` element fixes the issue.

I think we need a way to check whether the checkbox is displayed and add or omit that class depending on that.

farridav:
Hmm, ive removed the styles for `djn-checkbox-select-all` in main.css and it seems to render fine for both read only and normal users .. 

I'm not sure why that CSS was added in the first place, it may be just as easy as us removing it..

Blaza:
Yeah, if I override main.css and remove the styles it seems to work as expected.

If there is no need for that line in the CSS, maybe we should remove it then?

PR #412, opened 2022-06-02
Fix changelist results first column css class

akamoroz:
Fixed https://github.com/farridav/django-jazzmin/issues/362

The first column in changelist results don't have to contain checkbox.
When django admin actions are deactivated, the class `djn-checkbox-select-all` is not needed.
When the column has checkboxes the class is always `action-checkbox-column`:
https://github.com/django/django/blob/stable/3.2.x/django/contrib/admin/templatetags/admin_list.py#L90

Before fix:
Screenshot 2022-06-02 at 14 07 07

After fix:
Screenshot 2022-06-02 at 14 09 37


farridav:
LGTM

Issue #210, opened 2020-11-07
duplicate drop-downs when using django-autocomplete-light


frnmst:
Hi,

I am using [django-autocomplete-light](https://github.com/yourlabs/django-autocomplete-light) to display foreign key drop-down menus. This is what I get when using django-jazzmin:

![a](https://user-images.githubusercontent.com/8964425/98453289-d25b4400-2157-11eb-9ab2-291089e426ad.png)

The menu on the left is from jango-autocomplete-light.

Standard admin:

![b](https://user-images.githubusercontent.com/8964425/98453294-e141f680-2157-11eb-83a8-3afb2b26a278.png)

Any ideas to hide django-jazzmin's menus when django-autocomplete-light menus are present?

Thank you

farridav:
I think we fix this by ignoring the select that autocomplete light is using in main.js

On a side note, unless your using autocomplete light for something advanced, or using a very old version of Django, auto complete on foreign key fields is available from Django https://docs.djangoproject.com/en/2.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.autocomplete_fields

frnmst:
Thanks, I didn't know about ModelAdmin.autocomplete_fields! I'll give it a go.

djpretzel:
I had this issue as well & it was a bit confusing. I do think DAL exposes more of the underlying functionality of Select2 and can be useful in scenarios where the default Django autocomplete isn't as configurable, so compatibility would be awesome.

farridav:
If someone is able to tell me a CSS class that is applied to the select, we can add it here https://github.com/farridav/django-jazzmin/blob/master/jazzmin/static/jazzmin/js/change_form.js#L108

And this will stop us running select2 over it..

marcoooo:
Is there any work around available for now? I struggle to find the required css to hide the right one :-)

desarrollosrosarinos:
I'm having the same problem. I cannot use django automcomplete because is very slow when you have many foreign keys and inlines.
Is something we can do to make it work till you fix it?
Thank you, this theme looks great.

desarrollosrosarinos:
I fix it commenting this function in change_form.js
function applySelect2() {
        // Apply select2 to any select boxes that don't yet have it
        // and are not part of the django's empty-form inline
        //const noSelect2 = '.empty-form select, .select2-hidden-accessible, .selectfilter, .selector-available select, .selector-chosen select';
[...]

PR #413, opened 2022-06-07
fix duplicated select2 on dal

MathiasKowoll:
Add class to fix duplicated on auto-complete-light package. need to add on the autocomplete form the attribute as follows: attrs={'class': 'noSelect2DAL'}


MathiasKowoll:
#210

PR #418, opened 2022-07-01
Fix stacked inline admin not using `has_delete_permission`

### LINKED ISSUE #417: Stacked inline admin template does not check `has_delete_permission`
The stock admin inline templates in Django (stacked as well as tabular) check both `inline_admin_formset.formset.can_delete` as well as `inline_admin_formset.has_delete_permission`. In jazzmin this is different. In jazzmin, the stacked template only checks `can_delete`, the tabular template checks both. While this is not an issue in plain Django from what I can tell (because it sets `can_delete` based on `has_delete_permission`) it causes an issue when using [django_reverse_admin](https://github.com/daniyalzade/django_reverse_admin), as it does not set `can_delete`.

Jazzmin should check both `can_delete` as well as `has_delete_permission`, just like the stock templates do.

diesieben07:
Fixes #417

### DIFF
--- a/jazzmin/templates/admin/edit_inline/stacked.html
+++ b/jazzmin/templates/admin/edit_inline/stacked.html
@@ -36,7 +36,7 @@
                                     </a>
                                 {% endif %}
                             </h3>
-                            {% if inline_admin_formset.formset.can_delete and inline_admin_form.original %}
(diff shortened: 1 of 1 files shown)


farridav:
Good spot..

PR #420, opened 2022-07-01
Fix wrong for attribute on labels in fieldset.html

### LINKED ISSUE #419: Labels in stacked inline admin fieldsets have the wrong `for` attribute
The overridden Fieldset template in jazzmin does not correctly generate the `for` attribute for labels when it is used for an inline admin.
The result is, that the labels are not associated with their corresponding form fields and thus cannot be clicked.

diesieben07:
Fixes #419


farridav:
Nive

PR #424, opened 2022-07-15
Add auto_now_add to last_print field so it's created

dhvcc:
There was a bug in `test_app` which crashed creation of a book since `last_print` field was **_not nullable_**, but had no default value.

Error showcase

error


farridav:
Thanks

PR #425, opened 2022-07-15
Update field CSS

dhvcc:
# The problem

So there were some css declarations that were searching for `field-{type}` classes, but in Django we're using `field-{field_name}`. So that's why there were no default styles for fields not having any `v...Field` class attached to them (when there's a default widget for, like `IntegerField` which adds `vIntegerField`).
I removed those parts and now things like `DecimalField`, `GenericIPAddressField` and others are rendered with correct styles.

I also removed selects from `display: block` and `width: 100%` rule, becuase it was going out of bounds.

## Testing

As for testing, here's the model
```python
# Fields are chosen from the diff. Names are chosen to not conflict with types
class TestModel(models.Model):
    ip_field = models.GenericIPAddressField("IP")
    duration_field = models.DurationField("Duration")
    boolean_field = models.BooleanField("Null boolean", null=True)
    file_path_field = models.FilePathField("File path", path="./")
    file_field = models.FileField("File path")
```

### Django 2.2
**Before the change**
image
**After**
image

### Django 3.0
**Before**
image
**After**
image

### Django 4.0 (not even allowing NullBooleanField anymore)
**Before**
image
**After**
image

### Select width (all of the other fixed are already applied)
**Before**
Screenshot from 2022-07-15 18-13-43
**After**
Screenshot from 2022-07-15 18-15-35

### DIFF
--- a/jazzmin/static/jazzmin/css/main.css
+++ b/jazzmin/static/jazzmin/css/main.css
@@ -205,11 +205,8 @@ table.dataTable thead .sorting_desc_disabled::after {
 .form-group div .vLargeTextField,
 .form-group div .vURLField,
 .form-group div .vBigIntegerField,
-.form-group div.field-duration input[type="text"],
-.form-group div.field-identifier input[type="text"],
-.form-group div.field-generic_ip_address input[type="text"],
-.form-group .field-null_boolean select,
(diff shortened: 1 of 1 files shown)


farridav:
Nice work !

PR #431, opened 2022-08-02
Add more ru translations

dhvcc:


dhvcc:
Merged #432 to this branch so I can show that this doesn't break anything (or at least now any written test)

PR #432, opened 2022-08-05
Fix builds

dhvcc:
### 1. Revert #420

`{{ field.field }}` can be a dictionary which won't have `id_for_label`. You can see it in here https://github.com/farridav/django-jazzmin/blob/0c1d115f10ebcab347942cadc78e67d73340e2f6/tests/test_app/library/books/admin.py#L26 we use `__str__` as a part of the display which will result in django trying to lookup `id_for_field` from a dict.
image

#### Note

I didn't notice any issues in manual testing, so this may be an issue with the test (although I also doubt it). I think it's safe to revert until we can fix it or fix the test.

### 2. Update tox's config to fix issues with `conda`/`pyenv`

Update `tox` section in `pyproject.toml` to fix `InterpreterNotFound` error. Check out [solution reference](https://github.com/tox-dev/tox/issues/378#issuecomment-413937359) (note that I'm not using `conda`, but `pyenv`).

This was the error:
```
└❯ tox
django2-py36 create: /home/dhvcc/PycharmProjects/django-jazzmin/.tox/django2-py36
ERROR: InterpreterNotFound: python3.6
django2-py37 create: /home/dhvcc/PycharmProjects/django-jazzmin/.tox/django2-py37
ERROR: InterpreterNotFound: python3.7
django3-py36 create: /home/dhvcc/PycharmProjects/django-jazzmin/.tox/django3-py36
ERROR: InterpreterNotFound: python3.6
django3-py37 create: /home/dhvcc/PycharmProjects/django-jazzmin/.tox/django3-py37
ERROR: InterpreterNotFound: python3.7
```
This seems to not be the case for `gh-actions` as they're using different containers for each python version

### 3. Update `render_counts` and `templates_used` for `django>4`

Not sure what made this fail honestly as we can see it **worked** before, but I had persistent fails with running UI tests for `django>4` (also shown in recent `gh-actions` build). So yeah, I just updated the values accordingly and now it seems to be working for me locally and for github actions runs.

### DIFF
--- a/jazzmin/__init__.py
+++ b/jazzmin/__init__.py
@@ -1,4 +1,3 @@
-
 import django
 
 version = "2.5.0"
--- a/jazzmin/templates/admin/includes/fieldset.html
+++ b/jazzmin/templates/admin/includes/fieldset.html
@@ -22,7 +22,7 @@
     <div class="form-group{% if line.fields|length_is:'1' and line.errors %} errors{% endif %}{% if not line.has_visible_field %} hidden{% endif %}{% for field in line %}{% if field.field.name %} field-{{ field.field.name }}{% endif %}{% endfor %}">
         <div class="row">
             {% for field in line %}
-                <label class="{% if not line.fields|length_is:'1' and forloop.counter != 1 %}col-auto {% else %}col-sm-3 {% endif %}text-left" for={{ field.field.id_for_label }}>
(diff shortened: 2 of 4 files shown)


dhvcc:
It seems like some build are still failing, I'll update this one when have more free time

Issue #429, opened 2022-07-29
Please support fontawesome's python package~


HenryJi529:
Please check the link below~
https://fontawesome.com/docs/web/use-with/python-django

Issue #415, opened 2022-06-16
Upgrade to FontAwesome 6


adambirds:
It would be great if you could upgrade to FontAwesome 6 as there are a lot of new icons we can't use.

Thanks

PR #436, opened 2022-08-18
Migrating from local installed FontAwesome 5 to python module FontAwesome 6

AlphaKiloHotel:
Hello,
First contribution, please be kind 😆 

Moving project from using a locally installed FontAwesome 5 in static to FontAwesome 6 using the official python module "fontawesomefree". This way, FontAwesome could be used by the developper of the project outside of Django-jazzmin. So it sounds a great solution for me.

See Issues #415 and #429


dhvcc:
Hi, this seems like a good idea. It also seems that FA6 is backwards comtabile, so should be no big deal upgrading it.

pyproject.toml
```diff
@@ -39,6 +39,8 @@ exclude = ["tests", "docs"]
 [tool.poetry.dependencies]
 python = ">=3.6.2"
 django = ">=2.2"
+dj-database-url = "^1.0.0"
```
dhvcc, line 42: Are you sure this should be here?

docs/bugs_and_features.md
```diff
@@ -28,15 +28,13 @@ We welcome new features, here is a list of guidelines to consider:
...
-We prefer to use HTML first, failing that, use CSS, failing that, use JavaScript (This approach helps with maintainability) 
-
+We prefer to use HTML first, failing that, use CSS, failing that, use JavaScript (This approach helps with maintainability)
 
 Some useful links for feature development:
 
 - [https://adminlte.io/themes/v3/index3.html](https://adminlte.io/themes/v3/index3.html)
 - [https://adminlte.io/docs/3.0/index.html](https://adminlte.io/docs/3.0/index.html)
-- [Font awesome 5.13.0 free icons](https://fontawesome.com/icons?d=gallery&m=free&v=5.0.0,5.0.1,5.0.10,5.0.11,5.0.12,5.0.13,5.0.2,5.0.3,5.0.4,5.0.5,5.0.6,5.0.7,5.0.8,5.0.9,5.1.0,5.1.1,5.2.0,5.3.0,5.3.1,5.4.0,5.4.1,5.4.2,5.13.0,5.12.0,5.11.2,5.11.1,5.10.0,5.9.0,5.8.2,5.8.1,5.7.2,5.7.1,5.7.0,5.6.3,5.5.0,5.4.2)
+- [Font awesome 6.1 free icons](https://fontawesome.com/v6/search?m=free)
```
dhvcc, line 39: Glad we can get rid of this monstrosity :laughing:

dhvcc (commented):
LGTM, the only thing I'd ask is maybe you can split commits with markdown format fixes into a separate commit? It'll be easier to review commits that way in future :slightly_smiling_face:

pyproject.toml
farridav: good point, this is a dependency for the test app, not for jazzmin itself, you can add it into the dev dependencies if you like, though the test app dependencies are tracked here https://github.com/farridav/django-jazzmin/blob/master/tests/test_app/requirements.txt so they can get picked up by the heroku deploy

tests/test_app/library/settings.py
```diff
@@ -17,6 +17,7 @@
 ALLOWED_HOSTS = ["*"]
 INSTALLED_APPS = [
     # Keep this above 'django.contrib.admin'
+    "fontawesomefree",
```
farridav, line 20: if we want this to work, and for djangos static file finder to find the fontawesome files, then it will require everybody that has jazzmin installed to append this to their `INSTALLED_APPS` ..  this stops jazzmin being a one line drop in package for spicing up the admin

farridav:
I like the idea of vendoring in the FA files using a python package, it helps with maintainability, but given that it relies on the `fontawesomefree` app to be installed into a users `INSTALLED_APPS` I'm not sure it fits with the simplicity of jazzmins installation ..

dhvcc:
It's just my opinion, obviously, but I think that's ok since I guess it should now be the preferred method of using FA in python/django. Also, since jazzmin admin already uses FA, then it'd be easier for the main app to also use it (in case it's a template based app)

AlphaKiloHotel:
I don't think that adding the app 'fontawesomefree' in the INSTALLED_APPS would add any challenge for installation. There are no special manipulation needed outside that, so it appears ok for me as well

aetaylor:
I have a FA6 Pro license - would I be able to configure Jazzmin to use this?

katiam2:
Hey there! 👋
Looking at this PR from 2022, it seems the decision is clear - we're keeping our installation config lean. While the change is interesting, two years of discussion have shown we're sticking to our minimalist approach. I think it's time to close this one. Thanks for understanding! 🙌

PR #440, opened 2022-09-13
django-autocomplete-light full support

Simanas:
.noSelect2DAL solution requires using attrs={'class': 'noSelect2DAL'} with every autocomplete widget declaration. It is not  intuitive or beautiful in any way. 

This patch fully fixes issue mentioned on issue #210


farridav:
Great, this is the fix i was looking for

PR #446, opened 2022-10-11
Added multiple search_model setting option

jrdabrio:
Hi hi :smile:

Well, i have created this PR because I think it might be a good idea to have the posibility to add different models to `search_model` setting. Updating only a few things we have a new cool option!

image


jrdabrio:
Woops, this PR has an error, give me a sec

PR #447, opened 2022-10-11
Added multiple search_model setting option

jrdabrio:
Hi hi :smile:

Yes, this is te good one. Well, I have created this PR because I think it might be a good idea to have the posibility to add different models to `search_model` setting. Updating only a few things we have a new cool option!

image

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -36,8 +36,9 @@ JAZZMIN_SETTINGS = {
     # Copyright on the footer
     "copyright": "Acme Library Ltd",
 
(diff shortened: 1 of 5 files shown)


farridav:
This is a great addition, with minimal complexity introduced, and offers good value ..

Do you think you could update the documentation? so that jazzmin users know how to take advantage of this awesome feature ?

jrdabrio:
> This is a great addition, with minimal complexity introduced, and offers good value ..
> 
> Do you think you could update the documentation? so that jazzmin users know how to take advantage of this awesome feature ?

Sure! There you have it 😄

PR #450, opened 2022-10-21
Make use_google_fonts_cdn work with registration templates

alexkiro:
The `use_google_fonts_cdn` setting works great for the admin pages! However it seems that google fonts are still used in the registration pages (e.g. log out view, password reset view etc.). This PR should solve this issue.

Along the change itself, I've also updated the example setting documentation to reference this flag.


farridav:
Good spot!, Thanks for your contribution

Issue #479, opened 2023-02-10
Error in filters if they are related from same related model


Pijuli:
Hi!
First of all, thanks for the amazing job. Second, I'm sorry I can not make a PR, it's beyond my knowledge. 

Let's suppose we have the following models
![image](https://user-images.githubusercontent.com/1220004/218029469-645f2f16-fca3-4c78-ba38-aa2f291539f9.png)

and following filters
`list_filter = (
        ("host_contact", admin.RelatedOnlyFieldListFilter),
        ("host_contact__host", admin.RelatedOnlyFieldListFilter),
    )`

Two select2 will appear in the admin view and if you use them two query param will appear:
`[...]?host_contact__id__exact=5028&host_contact__host__id__exact=6771`

html from the select 2 before applying filters are:
host_contact select:
`data-name="host_contact__id__exact"`
host_contact__host select:
`data-name="host_contact__host__id__exact"`

After the first search, query params breaks and data's on the select are wrong
html from the select 2 after applying filters once are:
host_contact select:
`data-name="host_contact__host__id__exact"`
host_contact__host select:
`data-name="host_contact__host__id__exact"`

leading to having the following query params breaking the filtering until you clean them and having
`[...]?host_contact__id__exact=5028&host_contact__host__id__exact=6771`

And i can think of where the issue is, because both start with the same "host_contact..." string and field key 

https://github.com/farridav/django-jazzmin/blob/master/jazzmin/templatetags/jazzmin.py#L287

Tell me if this is helpful for you and makes any sense.

PR #480, opened 2023-02-10
Change value field match conditions in jazzmin_list_filter template tag

maltalk:
Hi,

This PR solves Issue #479

The problem appears when 2 list filters can match between them in  `key.startswith(field_key + "__")` condition.

In order to prevent this i added a condition to check on count of string "__" in both key and field_key variables.

I don't know if it's the best solution or even if works with all existing use cases as i don't know them all and couldn't find one where it goes wrong.


tura0 (approved):

JMSoler7 (approved):

n1ngu (approved):
The bugfix feels right to me but it is somewhat confusing. I'd really enjoy a unittest for the bug but I am clueless on how to write one. Sadly, few jazzmin tags feature a specific unit test.

PR #496, opened 2023-06-19
Jazzmin showing untranslated string while logging out

arneatec:
Jazzmin is showing untranslated 'Thanks for spending some quality time with the Web site today.' message in Django 4.x.x when doing standard logout (http://127.0.0.1:8000/bg/admin/logout/). Note the BG(Bulgarian language) in the url, uses internationalized urls.

The root of the problem is that jazzmin is using an outdated 'logged_out.html'. In it, the 'Thanks for..' string has a capital W in the word 'Web'. Django upstream has fixed this to use lowercase w ('web') in the following commit :
https://github.com/django/django/commit/7d909b2282c91c5bc5204dd160412e90fa86d624

As a result, jazzmin shows an untranslatable sting that is not present in either django or jazzmin .po files.

This pull request fixes the issue for the current Django release and future releases provided that the string does not mutate even further.


arneatec:
Please note that only the first commit (c372f5f8691a169b83dc2964677bfa25a4556ce9) is meant to be merged, the following one is to deal with the Github actions failing in the local repository.

PR #499, opened 2023-07-14
replace depcrecated command "length_is" with "length" Django 4 upwards

SourceDoctor:
replace depcreated length_is with length  for Django4

"
/root/.env/lib/python3.9/site-packages/django/template/defaultfilters.py:617: RemovedInDjango51Warning: The length_is template filter is deprecated in favor of the length template filter and the == operator within an {% if %} tag.
"

warning can be seen when running Django with "python -Wa manage.py runserver"

### DIFF
--- a/jazzmin/templates/admin/includes/fieldset.html
+++ b/jazzmin/templates/admin/includes/fieldset.html
@@ -19,16 +19,16 @@
 {% endif %}
 
     {% for line in fieldset %}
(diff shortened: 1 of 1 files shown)


FabianClemenz:
already resolved in [v.3.0.1](https://github.com/farridav/django-jazzmin/releases/tag/v3.0.1)

Issue #508, opened 2023-09-11
Warning errors for unset avatar field on user model


PavelPancocha:
We started getting a lots of "warning" errors like:

```
avatar field must be an ImageField/URLField on the user model, or a callable
```

I found out, it is because sometimes the "avatar" is not set, but it logs an error instead of just displaying the fallback icon.

PavelPancocha:
I did fix for it: https://github.com/farridav/django-jazzmin/pull/509

PR #509, opened 2023-09-11
Properly handle empty file field

PavelPancocha:
- Added some more code-coverage

Fix: #508

### DIFF
--- a/jazzmin/templatetags/jazzmin.py
+++ b/jazzmin/templatetags/jazzmin.py
@@ -178,15 +178,17 @@ def get_user_avatar(user: AbstractUser) -> str:
 
     # If we find the property directly on the user model (imagefield or URLfield)
     avatar_field = getattr(user, avatar_field_name, None)
-    if avatar_field:
-        if type(avatar_field) == str:
+    if avatar_field is not None:
+        if not avatar_field:
+            return no_avatar
+        elif type(avatar_field) == str:
             return avatar_field
         elif hasattr(avatar_field, "url"):
             return avatar_field.url
         elif callable(avatar_field):
             return avatar_field()
 
-    logger.warning("avatar field must be an ImageField/URLField on the user model, or a callable")
+    logger.warning("Avatar field must be an ImageField/URLField on the user model, or a callable")
 
     return no_avatar
 
--- a/tests/test_templatetags.py
+++ b/tests/test_templatetags.py
@@ -1,5 +1,5 @@
 import json
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, NonCallableMock
 
 import pytest
 from django.contrib.admin.models import CHANGE, LogEntry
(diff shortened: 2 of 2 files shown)


PavelPancocha:
Local tests:
```
====================================================================================================================== short test summary info =======================================================================================================================
FAILED tests/test_admin_views.py::test_detail - AssertionError: assert {'admin/base....html': 1, ...} == {'admin/base....html': 1, ...}
=================================================================================================================== 1 failed, 45 passed in 15.31s ====================================================================================================================
```

tests/test_templatetags.py
```diff
@@ -49,16 +49,47 @@ def test_style_bold_first_word():
...
     ],
 )
-def test_get_user_avatar(case, test_input, field, expected, custom_jazzmin_settings):
+def test_get_user_avatar(case, test_input, field, expected, log, custom_jazzmin_settings, caplog):
     """
     We can specify the name of a charfield or imagefield on our user model, or a callable that receives our user
     """
     custom_jazzmin_settings["user_avatar"] = field
     assert jazzmin.get_user_avatar(test_input) == expected
+    print(caplog.text)
```
jamesgilmorelyst, line 91: Can the print message be deleted, please?

jamesgilmorelyst (commented):
Thanks for the contribution. Looks to me a good change once you fix the test.

PavelPancocha:
@jamesgilmorelyst You took over the Jazzmin? Will you be able to merge it if it works?

PavelPancocha:
@jamesgilmorelyst @farridav please comment, that I know it is not a lost of time to revive this. Thanks!

tests/test_templatetags.py
farridav: Once done there's no reason we can't get this in .

PacificGilly: @PavelPancocha You also need to rebase. I've QA'd this now and LGTM to be apart from that you still have `dj_database_url` in the test_app. So can you rebase and we can look into merging. Thanks for your help on this.

tests/test_templatetags.py
```diff
@@ -49,16 +49,47 @@ def test_style_bold_first_word():
...
+    if log:
+        assert log in caplog.text
+    else:
+        assert not caplog.text
```
PacificGilly, line 95: I do love it when I see the logs being tested as well! 

Optional - Whats your thoughts on testing log levels (i.e. `WARNING` here).

jazzmin/templatetags/jazzmin.py
```diff
@@ -178,15 +178,17 @@ def get_user_avatar(user: AbstractUser) -> str:
 
     # If we find the property directly on the user model (imagefield or URLfield)
     avatar_field = getattr(user, avatar_field_name, None)
-    if avatar_field:
-        if type(avatar_field) == str:
+    if avatar_field is not None:
+        if not avatar_field:
+            return no_avatar
```
PacificGilly, line 183: Optional - `no_avatar` is basically a constant so potentially could be moved outside the function to the top of the file (or to the settings).

PacificGilly (approved):
LGTM once rebased - comments are optional. Thanks.

PavelPancocha:
@PacificGilly @farridav Ready :)

PavelPancocha:
Just Ruff failed for python 3.12 -> it could be formatted differently, but that doesn't work for python 3.8 - so I decided to keep it python 3.8 compatible. Maybe it could be set to always format it for lowest supported version?

farridav:
This looks like it's failing on the coveralls job, which is the only one that appears to be running ruff over the tests folder.. so it seems the formatting issue is in there..

Are you sure this build cannot be fixed ?

PavelPancocha:
@farridav I can remove the typing.

PavelPancocha:
@farridav I fixed it, now fails with:

```
Not on TravisCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var
```

farridav:
Nice.. yes this particular error is with me, I need to make it so that the secret is available for other people.. but that doesn't block this PR as it will work fine in main.

Thanks for your contribution

jcgiuffrida:
For anyone else who is still dealing with this warning after upgrading to 3.0.0, your `user_avatar` can't be a field on a related model because it'll silently fail to fetch, then warn you that you have no avatar field. Use a `@property` on your User model to reference the avatar instead.

PR #514, opened 2023-10-17
I18n: Adds translation strings for the pt_BR language

leandrodesouzadev:
This PR adds translations for the Brazilian Portuguese (pt_BR) language.


leandrodesouzadev:
Hey there! Love the project so far.
I haven't found the documentation on how to generate the translation strings. So i used the command: `poetry run django-admin makemessages -l pt_BR` and afterwards: `poetry run python cli.py locales --prune pt_BR`. With that, i noticed that my translation strings file is way bigger than the other languages. If there's any problems on it, i would be happy to fix it.

farridav (approved):

PR #526, opened 2024-01-05
Remove *mo blobs from source control

### LINKED ISSUE #505: Sporadic Execution of msgfmt failed 'msgid' and 'msgstr' entries do not both begin with '\n'
Running `django-admin compilemessages` will most usually decide the fr translation is up to date
```
File “...../jazzmin/locale/fr/LC_MESSAGES/django.po” is already compiled and up to date
```
but somteimes django will want to process it for some reason. And it will fail:
```
Execution of msgfmt failed: ..../jazzmin/locale/fr/LC_MESSAGES/django.po:88: 'msgid' and 'msgstr' entries do not both begin with '\n'
```

I am still puzzled about what triggers this but my builds are now sort of flaky (workaround is to simply re-trigger the build)

I guess this is related to https://www.gnu.org/software/gettext/FAQ.html#newline. I am not sure I understand the rationale for that gettext behavior but similar issues (e.g. https://github.com/scanmem/scanmem/issues/205) have been fixed by enforcing newline parity between original and translated strings: https://github.com/farridav/django-jazzmin/blob/master/jazzmin/locale/fr/LC_MESSAGES/django.po#L90

n1ngu:
Compile them at build time with a hatchling custom hook (replacing poetry!)

Closes #505


Endzel:
Any chance for this to progress @n1ngu ? This issue is happening in some of my deployments.

hatch_build.py
```diff
@@ -0,0 +1,10 @@
+import typing as t
+
+import django.core.management
+
+from hatchling.builders.hooks.plugin.interface import BuildHookInterface
+
+
+class CustomBuildHook(BuildHookInterface):
+    def initialize(self, version: str, build_data: dict[str, t.Any]) -> None:
+        django.core.management.call_command("compilemessages")
```
n1ngu, line 10: To be clear, I am uncomfortable with the dependency introduced by this instruction. The only reason to do this was validating the workflow. But if community is ok with removing *.mo files and compiling them at wheel build time and replacing poetry with hatch, I'll invest some effort to rewrite this with babel or gettext.

PR #531, opened 2024-02-05
Added styles for RTL languages

daninouai:


ahmedalrifai:
@farridav Any news on this fix?

farridav:
There's a huge amount of duplication here, is there not a way we can "cascade" the style sheets in an additive fashion? I.e - only override that which needs overriding from the LTR version of the css ?

PR #532, opened 2024-02-13
feat: Make fieldset take all awailable space

last-partizan:
In some previous version of the django-jazzmin, this block was full-size, then it was changed to smaller size.

It's better for it to be full-size.


katiam2:
This is a duplicate of [#606 ](https://github.com/farridav/django-jazzmin/pull/606 ).

PR #542, opened 2024-03-22
logout fixed for django 5

arifogluisa:
logout doesn't work for Django 5.0 , cause logout view has changed from get method to post method


farridav:
This appears to work fine with all supported versions of Django, thanks for the contribution

farridav:
This has now been merged in as part of https://github.com/farridav/django-jazzmin/pull/544

PR #548, opened 2024-03-25
Improve static code checks

farridav:
- Adds type checks with mypy
- Adds types
- Fixes flake8 issues


jamesgilmorelyst:
Nice change. Great to have mypy in this repo!

Curious - should we look at adding `ruff` as well to this repo so we can remove, black, flake8 and isort? Happy to raise a PR for that.

PR #550, opened 2024-03-26
Allow the creating of new pypi packages from github releases alone

farridav:
This _should_ allow for us to create releases on github with zero code changes, allowing for a test, code, review, merge and release when ready from main approach

- `testpypi` (which i use to validate the package build and deploy works as expected) will continually bump patches on the latest published version
- pypi.org will receive the version of the tag created in github (with the `v` prefix removed)
- Package versions need not be tracked in the code at all

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -26,3 +26,16 @@ jobs:
           poetry version prepatch
           poetry config repositories.test_pypi https://test.pypi.org/legacy/
           poetry publish --build -r test_pypi --username __token__ --password ${{ secrets.TEST_PYPI_TOKEN }} || true
+
+  update_release_draft:
+    permissions:
+      contents: write
+      pull-requests: read
+    runs-on: ubuntu-latest
+    steps:
+      - uses: release-drafter/release-drafter@v5
+        with:
+          config-name: release-drafter.yml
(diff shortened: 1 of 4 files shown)


.github/workflows/pre-release.yml
```diff
@@ -21,5 +21,8 @@ jobs:
 
       - name: Deploy to testpypi.org
         run: |
+          LATEST_RELEASE=${curl "https://api.github.com/repos/farridav/django-jazzmin/tags" | jq -r '.[0].name'/#v}
```
PacificGilly, line 24: I've not managed to get this to work. I get a compile error. Maybe this is needed?
```suggestion
          LATEST_RELEASE=${curl -s "https://api.github.com/repos/farridav/django-jazzmin/tags" | jq -r '[0].name[1:]'}
```

.github/workflows/pre-release.yml
```diff
@@ -21,5 +21,8 @@ jobs:
...
+          poetry version $LATEST_RELEASE
+          poetry version patch
```
PacificGilly, line 26: Why patch? Should we release a prepatch before?
```suggestion
          poetry version prepatch
```

PacificGilly (approved):
Great change idea to have a new test release per PR merge.

PR #544, opened 2024-03-23
Logout issue

farridav:
Extends: https://github.com/farridav/django-jazzmin/pull/542

By:
- Fixing a failing test
- dropping support for Python 3.6
- Upgrading `pytest-cov`

PR #534, opened 2024-02-28
Add support for Django 5.1 by removing deprecated features

jamesgilmorelyst:
There is only one Django feature used here that is removed in >=5.0, which is the `length_is` filter query:

```
RemovedInDjango51Warning: The length_is template filter is deprecated in favour of the length template filter and the == operator within an {% if %} tag.
```

It is a trivial change to make, and the `length` feature if backwards compatible to the earliest support version this repo supports of 2.2.

Preview of `fieldsets`, which has been modified in this PR:
Screenshot 2024-02-28 at 18 22 25

As this is my first PR, I've also added in some QoL changes for new devs working in this repo. Hopefully that's okay 🙂

PR #551, opened 2024-03-29
Update Python & Django supported versions + migrate to ruff

samiashi:
Update Python supported versions - https://devguide.python.org/versions/
- Drop support for Python 3.6 & 3.7
- Add support for Python 3.11 & 3.12

Update Django supported versions - https://www.djangoproject.com/download/#supported-versions
- Drop support for Django 2.2, 3.0 & 3.1
- Add support for Django 4.2 & 5.0

Migrate to ruff - https://github.com/astral-sh/ruff
- Replace flake8 & black with ruff
- Format files with ruff

Miscellaneous
- Fixed markdown violations


tox.ini
```diff
@@ -0,0 +1,41 @@
+[tox]
```
PacificGilly, line 1: Curious - Why the move out of pyproject.toml?

pyproject.toml
```diff
@@ -36,7 +37,7 @@ exclude = ["tests", "docs"]
 
 [tool.poetry.dependencies]
 python = ">=3.8"
-django = ">=4.2,<5.0"
+django = ">=4.2"
```
PacificGilly, line 40: Further changes are needed in 5.1 of Django to maintain support - https://github.com/farridav/django-jazzmin/pull/534
```suggestion
django = ">=4.2,<5.1"
```

tests/test_admin_views.py
```diff
@@ -42,7 +42,7 @@ def test_logout(admin_client):
     """
     url = reverse("admin:logout")
 
-    response = admin_client.get(url)
+    response = admin_client.post(url)
```
PacificGilly, line 45: Interesting this wasn't already done, as I thought we fixed the logout issue already

PacificGilly (approved):
Amazing PR, and thanks for grouping these nicely in all the commits. Looks like you still need to rebase first though. Approving to trigger the CI to check this builds properly

tox.ini
```diff
@@ -0,0 +1,43 @@
...
+envlist =
+    django4-{python3.8,python3.9,python3.10,python3.11,python3.12},
+    django5-{python3.10,python3.11,python3.12},
+    # run one of the tests again but with coverage
+    coveralls-django4-python3.11,
```
PacificGilly, line 6: Should we use the latest PY version for the coveralls test? cc: @farridav 
```suggestion
    coveralls-django4-python3.12,
```

PacificGilly:
@farridav the 3.11 build is failing
https://github.com/farridav/django-jazzmin/actions/runs/8481330654/job/23245748961?pr=551
```
  coveralls.exception.CoverallsException: Not on TravisCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
```
I'm not sure why it can't find the `COVERALLS_REPO_TOKEN` env var as your PRs work without issue. Without seeing the settings, I'm guessing this token is also working with your account maybe?
https://github.com/samiashi/django-jazzmin/blob/8a20f52fed70ee8de7a80ae7c2a9d093f3a09830/.github/workflows/build.yml#L18

tox.ini
farridav: I agree, I'd rather it stay in project.toml, despite It not being formatted so nicely there, part of the reason in so fond of the PEP for it is that it reduces the amount of tooling specific files in repo roots

.gitignore
```diff
@@ -19,3 +19,5 @@ admindocs_templates
 *.sqlite3
 tests/test_app/static/
 .python-version
+.mypy_chache
```
farridav, line 22: Cache ?

.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
+        python-version: [3.8, 3.9, '3.10', 3.11, 3.12]
```
farridav, line 16: +1 :)

jazzmin/__init__.py
```diff
@@ -1,6 +1 @@
-import django
-
-version = "2.6.1"
-
-if django.VERSION < (3, 2):
-    default_app_config = "jazzmin.apps.JazzminConfig"
+version = "3.0.0"
```
farridav, line 1: Versions no longer tracked here, we can set this to 0.0.0

pyproject.toml
```diff
@@ -80,13 +75,29 @@ norecursedirs = ".git .tox _resource .mypy_cache build dist docs"
...
+ignore = [
+    "E501", # line too long
+]
+
+[tool.ruff.lint.isort]
+known-first-party = ["jazzmin"]
+known-local-folder = ["tests"]
+
+[tool.ruff.lint.mccabe]
+max-complexity = 12
```
farridav, line 100: 10 is a more sensible number, I find that it catches things that are genuinely in need of complexity reduction

pyproject.toml
```diff
@@ -80,13 +75,29 @@ norecursedirs = ".git .tox _resource .mypy_cache build dist docs"
...
+select = [
+    "E",    # PEP8 errors
+    "W",    # PEP8 warnings
+    "F",    # PyFlakes
+    "I",    # isort
+    "C",    # mccabe
+    "T201", # flake8-print
+]
+ignore = [
+    "E501", # line too long
```
farridav, line 92: Why would we want to ignore this ? Surely ruff will format it all to our line length of 120 anyway

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "django-jazzmin"
-version = "2.6.1"
+version = "3.0.0"
```
farridav, line 3: This can also be 0.0.0, as versions are tracked outside, allowing us to quantify major/minor/patch at the point of release as opposed to the point of change.

pyproject.toml
```diff
@@ -65,7 +56,11 @@ factory-boy = "^3.2.1"          # Factory generation
 mypy = "^0.931"                 # Type checking
 click = "^8.0.3"                # Framework for building cli's
 ipdb = "^0.13.9"                # ipython breakpoints
-black = "^21.12b0"              # Code formatting and linting (with pyproject.toml support)
+ruff = "*"                      # linter and code formatter
```
farridav, line 59: Loving this

pyproject.toml
farridav: Hmm, if we do this, then people won't even bother installing the app beyond this version, nor will they be able to tolerate or fix minor issues in unsupported versions.

The decision to have an upper bound is contentious I admit, if this package is absolutely broken on Django 5.1 then maybe it's a good idea, but otherwise, maybe we should leave it open and endeavour to resolve issues as they arise

farridav (changes requested):
Some great changes here, and will help us with maintenance.

jazzmin/__init__.py
PacificGilly: Being fixed in this PR to remove the hardcoding

pyproject.toml
PacificGilly: Ah yeah good point. It's not totally broken in 5.1. This seems to only impact fieldsets, so likely still a widespread issue.

farridav:
Yeah it's odd, probably some sort of a permissions issue, I will see if I can make sense of it and get the coveralls token working properly

tox.ini
samiashi: Makes sense 👍🏻 
Moved back to pyproject.toml

.gitignore
samiashi: 🤦🏻‍♂️ 
Updated 👍🏻

pyproject.toml
samiashi: Updated 👍🏻

pyproject.toml
samiashi: Updated 👍🏻

tests/test_admin_views.py
samiashi: The logout via GET was deprecated in Django 4.1 hence why it wasn't earlier
https://docs.djangoproject.com/en/4.2/releases/4.1/#log-out-via-get

PacificGilly: Yeah, I think it might have been missed in https://github.com/farridav/django-jazzmin/pull/544

farridav (approved):

farridav:
Right, lets get this moving.. a massive step in the right direction here .. 

Next steps will be to make the entire codebase properly typed, start making use of pydantic (or at least data classes) for configuration.

And get some better test coverage :)

farridav:
Can we get a reabase on this ? Then we can get it merged ...

samiashi:
> Can we get a reabase on this ? Then we can get it merged ...

@farridav Should be rebased with master now

PR #545, opened 2024-03-23
Start updating version support for python/django

farridav:
- Support Python 3.8 -3.12
- Support Django 3-5
- Update all dependencies to match new version constraints

TODO:
- [ ] Make required fixes to support new versions

PR #555, opened 2024-03-29
Fix displaying the installed version of Django-Jazzmin in the UI. 

PacificGilly:
Since we changed our CD process, and it now will automatically generate test
builds (https://github.com/farridav/django-jazzmin/pull/550), the version in the `__init__` file now needs to be
updated to stop it from being hard coded. As in development, the version
is now set to 0.0.0 in the `pyproject.toml` file; we need to get the
version dynamically. We can call the package manager to check which
version of `django-jazzmin` was installed. This, sadly, does vary
depending on which version of Python is installed - support for these
older versions is soon to be dropped.

### In Developent
Screenshot 2024-03-29 at 19 01 33

### When Installed 
```
In [1]: from importlib.metadata import version
   ...:
   ...: version = version("django-jazzmin")

In [2]: version
Out[2]: '2.6.2a0'
```

Also a few minor fixes to `black` and `poetry` in pyproject.toml I came across.


farridav (approved):

farridav:
3.7 is not enjoying these approaches to getting installed package versions..

PacificGilly:
> 3.7 is not enjoying these approaches to getting installed package versions..

Yeah sadly, I'll do some more digging to see what I can do, or might have to wait for #545 or #551 to drop support for 3.6/3.7

PacificGilly:
I've added `django-jazzmin` to the tox deps list as all versions of Python where failing tox when testing locally.

PacificGilly:
Merging as all tests pass, with only `coveralls-django4-python3.10` failing due to the known issue with the `COVERALLS_REPO_TOKEN` env-var.

PR #565, opened 2024-04-18
Fix again the strategy for determining the latest version released.

PacificGilly:
Since we want to auto-publish the a test version of Django Jazzmin each time a PR is merged, we have to take into account the following:
* You need to know which test versions have already been released on test PyPi so, between production releases, you don't have tag clashes.
* When a production version is released to PyPi you need to start releasing test version using the new production version. Otherwise you'll stay on the initial version constantly.

### Before
| Test Version | Production Version |
|--------|--------|
| 2.6.2a0 | 2.6.2 |
| 2.6.2a1 | 2.6.2 |
| 2.6.2a1 | 2.7.0 | # Release to production
| 2.6.2a2 | 2.7.0 |

### After
| Test Version | Production Version |
|--------|--------|
| 2.6.2a0 | 2.6.2 |
| 2.6.2a1 | 2.6.2 |
| 2.6.2a1 | 2.7.0 | # Release to production
| 2.6.7a0 | 2.7.0 |

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -21,8 +21,16 @@ jobs:
 
       - name: Deploy to testpypi.org
         run: |
-          # Using `pypi` rather than our own tags as we don't create a "release" for test version of our package.
-          LATEST_RELEASE=$(curl -s https://test.pypi.org/rss/project/django-jazzmin/releases.xml | sed -n 's/\s*<title>\([{a,b}0-9.]*\).*/\1/p' | head -n 2 | xargs)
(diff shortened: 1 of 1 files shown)


farridav:
Quite complex, but I understand the rationale, and it seems to behave as expected.

PacificGilly:
> Quite complex, but I understand the rationale, and it seems to behave as expected.

Yeah, sadly not an elegant solution, but does the trick. Did you have a better solution or are you happy for me to merge?

farridav:
No let's get this moving, thanks for the work on it

farridav (approved):

PR #568, opened 2024-04-22
Fix Deployment Test Syntax bug

PacificGilly:
I guess for this exact reason is why it will be useful to have this `pre-release.yml` file being ran on PRs to validate nothing gets broken before merging :)


PacificGilly:
@farridav let me know what you think of this deployment change to ensure less issues after merging to master.

There is a strange issue with the `Deploy to testpypi.org (Dry Run)` step:
```
The "--password" option requires a value
```
https://github.com/farridav/django-jazzmin/actions/runs/8782873944/job/24097897693

I don't really understand why this is occurring; it is potentially caused by secrets that are not visible to PRs.

PacificGilly:
Nevermind, this seems to be by design - https://securitylab.github.com/research/github-actions-preventing-pwn-requests/. Makes sense now I've read this 😅

Anyway, I'll just remove this last check as the others are good enough.

farridav (approved):

PR #606, opened 2024-11-13
Fieldset uses col-9 not col-7

longhotsummer:
Otherwise the full width is not used, producing a cramped layout.


farridav:
coveralls check failing, will resolve this in a follow up PR, thanks for the contribution !!

longhotsummer:
Thanks!

PR #611, opened 2024-11-27
Update references for the master branch to become main

katiam2:
After master branch was renamed to become main, PR checks stopped running. And some of the docs now have incorrect links. This PR updates references for master branch.


farridav:
Good spot !

PR #634, opened 2025-04-25
Update base.html

AnesIDAW:
- Customize navbar.

Put navbar inside a block "{% block top_navbar %} {% endblock %}" so it can be customized when extends "admin/base.html".


AnesIDAW:
I am building a Django project and using jazzmin UI for my templates, when I tried to customize the navbar I figure out that is not enclosed in a block, so I edit the admin base.html to handle that, the template nav bar looks like that:
image

PR #638, opened 2025-07-03
Update main.css: set width/height for the GeoDjango OSM widget

### LINKED ISSUE #610: map is not showing
I'm using jazzmin with django admin, and GISModelAdmin should be displayed as a map, 
but it's not displayed because this error occurs.

"No map visible because the map container's width or height are 0."

It's displayed well in the basic admin without using jazzmin. 
I think there's a problem with jazzmin. How do I solve it?

sposs:
It's necessary to have in the CSS the width and height of the map object. Sure it could be added as local tweak, but it's a base django thing.

Fixes #610

### DIFF
--- a/jazzmin/static/jazzmin/css/main.css
+++ b/jazzmin/static/jazzmin/css/main.css
@@ -925,3 +925,10 @@ body.jazzmin-login-page {
     float: left;
     width: 100%;
 }
+
+/* GIS MAPS */
+.dj_map {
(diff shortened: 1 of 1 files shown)


skidipap:
yeah, it solved geodjango admin in jazzmin. love the patch. hope will merge soon

sposs:
Given that this pull request has been hanging there for months now, I think the repo is dead...

PR #639, opened 2025-07-08
Fixed inline tabs requiring a page refresh in change forms

BigDids:
## This fixes a known issue (#642) where a page refresh was required after clicking a tab in change forms for its content to show. The carousel tab display is partially fixed ; the fieldset label doesn't update when changing tabs.

_Note : This branch also contains a new feature that I requested a few months ago to add filters into the dashboard page and the side menu (as seen below)._

___

## Added two inputs in the side menu and dashboard page to filter models dynamically with JS.
The rendering of these inputs can be disabled in Jazzmin configuration: `show_model_filters`.

Empty inputs :

no_filters

Filtered models :

dashboard_and_sidebar_filters

### DIFF
--- a/jazzmin/static/jazzmin/css/main.css
+++ b/jazzmin/static/jazzmin/css/main.css
@@ -159,7 +159,7 @@ table.dataTable thead .sorting_desc_disabled::after {
 }
 
 .select2-container {
-    min-width: 300px;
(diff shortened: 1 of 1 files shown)


BigDids:
May I have an update on this request ? Thank you

Issue #674, opened 2026-03-22
collectstatic still fails in 3.0.4 — missing bootstrap.bundle.min.js.map


jskitz:
## Description                                                                                                                                                                                              
   
  The fix in 3.0.3 for #651 addressed the missing Bootswatch .map files and added bootstrap.min.js.map, but missed bootstrap.bundle.min.js.map. This means collectstatic still fails for any project using 
  ManifestStaticFilesStorage or WhiteNoise's CompressedManifestStaticFilesStorage.
                                                                                                                                                                                                           
## Error                      

  Post-processing 'vendor/bootstrap/js/bootstrap.bundle.min.js' failed!
                                                                                                                                                                                                           
  whitenoise.storage.MissingFileError: The file 'vendor/bootstrap/js/bootstrap.bundle.min.js.map'                                                                                                          
  could not be found with <whitenoise.storage.CompressedManifestStaticFilesStorage object at 0x...>.                                                                                                       
                                                                                                                                                                                                           
  The JS file 'vendor/bootstrap/js/bootstrap.bundle.min.js' references a file which could not be found:                                                                                                    
    vendor/bootstrap/js/bootstrap.bundle.min.js.map                                                                                                                                                        
                                                                                                                                                                                                           
## Root Cause                 
                                                                                                                                                                                                           
  jazzmin/static/vendor/bootstrap/js/bootstrap.bundle.min.js (Bootstrap 5.3.3) contains:                                                                                                                   
   
  //# sourceMappingURL=bootstrap.bundle.min.js.map                                                                                                                                                         
                             
  But the corresponding .map file was not included in the package. Looking at the directory contents:                                                                                                      
   
  jazzmin/static/vendor/bootstrap/js/                                                                                                                                                                      
  ├── bootstrap.bundle.min.js      ← references bootstrap.bundle.min.js.map (MISSING)
  ├── bootstrap.min.js             ← references bootstrap.min.js.map (present ✓)                                                                                                                           
  └── bootstrap.min.js.map         ✓                                                                                                                                                                       
                                                                                                                                                                                                           
## Expected Fix                                                                                                                                                                                             
                             
  Either:
  1. Include the missing file: Add bootstrap.bundle.min.js.map to jazzmin/static/vendor/bootstrap/js/ (this is the standard Bootstrap 5.3.3 source map available from the Bootstrap distribution), or
[...]

PR #675, opened 2026-03-27
fix(static): include missing bootstrap.bundle.min.js.map for collectstatic

Rakanhf:
Fix missing `bootstrap.bundle.min.js.map` causing collectstatic failures Issue [674](https://github.com/farridav/django-jazzmin/issues/674)

Ref : https://getbootstrap.com/docs/5.3/getting-started/download/

Also this effect this test: https://github.com/farridav/django-jazzmin/blob/8b5fa3726fa86084ea4e433378f1bbc2c1c1e7a5/tests/test_collectstatic.py#L44-L53

which is also mentioned in github issue


davidCarlos:
@farridav Please, merge this PR. Without it, we cannot use django-jazzmin in production.

